PR #78, opened 2019-06-24
Start using Thin Provisioning instead of regular DM snapshots

luxas:
Will add more details later


luxas:
closing in favor of #79, which wraps this and adds more functionality

PR #79, opened 2019-06-27
Thin Provisioning, combined with Declarative Configuration

luxas:


luxas:
This is _very_ stale at this point, I'll close this PR.
Also, with the devicemapper snapshotter merged into containerd v1.3, we'll probably use that instead for this feature.

PR #92, opened 2019-07-02
Add the device mapper pool and devices from `config_v2`

twelho:


luxas:
Thanks.

For context, this code is not used yet in the current codebase, but will shortly.

PR #93, opened 2019-07-03
Serializer, Storage and Client implementations

luxas:
@twelho this shows how we could abstract away
a) the underlying storage (could provide other forms, e.g. using k8s API, etcd or GitOps),
b) how the filtering/matching works
c) caching for resources only partially loaded from disk

Note that pretty much all of the client is autogenerated based on a template,  that is nice :smile:

Let's use this ASAP for the ongoing refactor, and rebase the whole program on top of this approach.


luxas:
This is ready to merge :tada:

PR #94, opened 2019-07-03
Support new types with current metadata handling

twelho:
Transform the existing `imgmd`, `kernmd` and `vmmd` to use the new types in `v1alpha1`.
This doesn't add any functionality, just changes the metadata handling format.


twelho:
Compiles and tested with `ignite run weaveworks/ignite-ubuntu`.

PR #95, opened 2019-07-04
Move types from `metadata` to `meta/v1alpha1`, replace k8s types with Ignite types

twelho:
`types` now only depends on Ignite's `meta`, many types have been moved to `v1alpha1/meta` regarding e.g. time, size, UID and networking.


twelho:
The issue with the `Size` marshaling fixed itself, `attach` and `ps` work again

twelho:
There are some updates to the `Makefile`, which we might want to merge separately.

luxas:
Merging now, it's fine as-is

PR #103, opened 2019-07-08
Cleanup work for v0.4.0

twelho:
Small improvements and bugfixes for the `v0.4.0` release of Ignite.
- [ ] Cache reworked


pkg/gitops/gitops.go
```diff
@@ -65,7 +65,7 @@ func RunLoop(url, branch string) error {
 				})
 			case gitops.UpdateTypeDeleted:
 				go runHandle(wg, func() error {
-					return handleDelete(file.APIType)
+					return handleDelete(vm)
```
luxas: this will panic, as VM is not defined here; it is `nil` as it has been removed.
instead, construct a new VM object which you pass, based on the API type

pkg/operations/remove.go
```diff
@@ -2,6 +2,7 @@ package operations
 
 import (
 	"fmt"
+	"github.com/weaveworks/ignite/pkg/metadata/vmmd"
```
luxas: this is weird, move to the right location

pkg/storage/storage.go
```diff
@@ -2,13 +2,11 @@ package storage
...
 import (
 	"fmt"
-	"path"
-	"strings"
-
 	"github.com/weaveworks/ignite/pkg/apis/ignite/scheme"
 	meta "github.com/weaveworks/ignite/pkg/apis/meta/v1alpha1"
 	"github.com/weaveworks/ignite/pkg/constants"
 	"github.com/weaveworks/ignite/pkg/storage/serializer"
+	"path"
```
luxas: move this to the group with the other stdlib packages

api/openapi/openapi_generated.go
```diff
@@ -514,6 +514,12 @@ func schema_pkg_apis_ignite_v1alpha1_VMSpec(ref common.ReferenceCallback) common
 							Ref: ref("github.com/weaveworks/ignite/pkg/apis/meta/v1alpha1.Size"),
 						},
 					},
+					"networkMode": {
```
luxas: rebase this to latest master and see if these changes still apply or go away...

pkg/gitops/gitops.go
twelho: Fixed in ad0cc62.

pkg/operations/remove.go
twelho: Moved in 650dad1.

pkg/storage/storage.go
twelho: Moved in 650dad1.

api/openapi/openapi_generated.go
twelho: Rebased, those changes seem to have disappeared.

PR #106, opened 2019-07-08
v0.4.0 cleanup 2, fix TypeMeta and the new Cache

twelho:
This PR contains the reworked cache (which should be robust) and backs the `DefaultStorage` with it.
Also contains a temporary fix for the unpopulated `TypeMeta`.

### DIFF
--- a/pkg/storage/cache.go
+++ b/pkg/storage/cache.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	meta "github.com/weaveworks/ignite/pkg/apis/meta/v1alpha1"
+	"log"
 )
 
 // Cache is an intermediate caching layer, which conforms to Storage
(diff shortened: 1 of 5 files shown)


pkg/storage/cache.go
```diff
@@ -200,23 +229,63 @@ func (c *objectCache) storeAll(objs []meta.Object) {
...
 func (c *objectCache) delete(kind meta.Kind, uid meta.UID) {
 	if uids, ok := c.objects[kind]; ok {
 		delete(uids, uid)
 	}
 }
 
 func (c *objectCache) count(kind meta.Kind) uint64 {
-	return uint64(len(c.objects[kind]))
+	count := uint64(len(c.objects[kind]))
+	log.Printf("cache: counted %d %s objects\n", count, kind)
```
luxas: in a follow-up, might want to remove this

twelho: Yes, that's removed along with all other debug prints when I get to verifying all the functionality

PR #107, opened 2019-07-08
Object inspection support

twelho:
This PR adds the `ignite inspect` command to retrieve the information/configuration of an Ignite object.


cmd/ignite/cmd/inspect.go
```diff
@@ -0,0 +1,44 @@
...
+			}())
+		},
+	}
+
+	addInspectFlags(cmd.Flags(), i)
+	return cmd
+}
+
+func addInspectFlags(fs *pflag.FlagSet, i *run.InspectFlags) {
+	fs.BoolVarP(&i.YAMLOutput, "yaml", "y", false, "Output the object in YAML format instead of JSON")
```
luxas: doesn't have to be done now, but eventually this should be done as `--output=<type>` => `-o=yaml`, `-o=json` instead for reusability and consistency.

cmd/ignite/cmd/inspect.go
```diff
@@ -0,0 +1,44 @@
+package cmd
+
+import (
+	"github.com/lithammer/dedent"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/weaveworks/ignite/cmd/ignite/run"
+	"github.com/weaveworks/ignite/pkg/errutils"
+	"io"
```
luxas: should be in its own group up top

cmd/ignite/run/inspect.go
```diff
@@ -0,0 +1,61 @@
+package run
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/weaveworks/ignite/pkg/apis/ignite/scheme"
+	"github.com/weaveworks/ignite/pkg/client"
+	"github.com/weaveworks/ignite/pkg/filter"
```
luxas: nit: grouping

cmd/ignite/run/inspect.go
```diff
@@ -0,0 +1,61 @@
...
+	*InspectFlags
+	object meta.Object
+}
+
+func (i *InspectFlags) NewInspectOptions(k, objectMatch string) (*inspectOptions, error) {
+	var err error
+	var kind meta.Kind
+	io := &inspectOptions{InspectFlags: i}
+
+	switch k {
```
luxas: `strings.ToLower(k)` to support `ignite inspect Image foo` if somebody would do it. no need to error in that case

cmd/ignite/cmd/inspect.go
twelho: Reworked in c6622cf.

cmd/ignite/cmd/inspect.go
twelho: Fixed in c6622cf.

cmd/ignite/run/inspect.go
twelho: Fixed in c6622cf.

cmd/ignite/run/inspect.go
twelho: Yes, was not sure if it's needed as it could train the user "incorrect" usage. Implemented in c6622cf now.

PR #113, opened 2019-07-09
Add Ignite usage documentation

twelho:
`docs/usage.md` explains a lot about managing VMs and their lifecycle/resources with Ignite.


twelho:
How does the user get the `weaveworks/ignite:version` Docker image containing `ignite-spawn`? Is that being pulled automatically?

luxas:
> Is that being pulled automatically?

yes

PR #118, opened 2019-07-10
Add next-action for better developer-workflow

alexellis:
Adds a next-action of using Ignite after installing it.

Signed-off-by: Alex Ellis <alexellis2@gmail.com>

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -34,6 +34,8 @@ Ignite version: version.Info{Major:"0", Minor:"3", GitVersion:"v0.3.0", GitCommi
 Firecracker version: v0.16.0
 ```
 
+Now you can continue with the [Getting Started Walkthrough](./docs/usage.md).
+
 ## Removing the installation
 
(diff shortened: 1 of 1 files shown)


alexellis:
Thanks for merging!

@luxas should I need to install KVM packages on Ubuntu for instance? If so, I would like to add a link to the requirements docs page i.e. See also: KVM on Ubuntu / KVM on CentOS / KVM on Fedora.

luxas:
> should I need to install KVM packages on Ubuntu for instance

No, just KVM support in the kernel/processor is enough.
Is the current merging in https://github.com/weaveworks/ignite/blob/master/docs/installation.md#system-requirements ok?

alexellis:
Looks good. I wonder if you can install `kvm-ok` without installing a load of other packages for libvirt?

```
# kvm-ok

Supported...
```

luxas:
I just checked that yes, you can get it via a `apt-get install cpu-checker`, without libvirt deps. I'll create an issue

PR #125, opened 2019-07-10
improve README.md

seeekr:
I believe this makes the phrase a little more idiomatic English, though I'm sure a true native speaker could come up with something even better.


seeekr:
You're welcome, keep up the good work!

Issue #130, opened 2019-07-11
Can not import image 


junaid18183:
The import image command is failing. 

```
# ignite image import weaveworks/ignite-ubuntu:latest
INFO[0000] Starting image import...
failed to format image 5f3dc10d89792555: command ["mkfs.ext4" "-I" "256" "-E" "lazy_itable_init=0,lazy_journal_init=0" "/var/lib/firecracker/image/5f3dc10d89792555/image.ext4"] exited with "mke2fs 1.42.9 (28-Dec-2013)\n/var/lib/firecracker/image/5f3dc10d89792555/image.ext4 is not a block special device.\nProceed anyway? (y,n) ": exit status 1

ignite version - 
# ignite version
Ignite version: version.Info{Major:"0", Minor:"4", GitVersion:"v0.4.0", GitCommit:"e0e7e8c50cd3c6532486625393dff4d415081829", GitTreeState:"clean", BuildDate:"2019-07-10T12:30:11Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Firecracker version: v0.17.0

OS and kernal  version - 
# cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.6 (Maipo)
# uname -a
Linux ip-172-31-4-245.us-gov-west-1.compute.internal 3.10.0-957.10.1.el7.x86_64 #1 SMP Thu Feb 7 07:12:53 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux```

junaid18183:
Manual running the command fails with error invalid blocks 
```[root@ip-172-31-4-245 ~]# mkfs.ext4 -I 256 -F lazy_itable_init=0,lazy_journal_init=0 /var/lib/firecracker/image/5f3dc10d89792555/image.ext4
mke2fs 1.42.9 (28-Dec-2013)
mkfs.ext4: invalid blocks '/var/lib/firecracker/image/5f3dc10d89792555/image.ext4' on device 'lazy_itable_init=0,lazy_journal_init=0'```

junaid18183:
created a PR - https://github.com/weaveworks/ignite/pull/131

jiangpengcheng:
met the same problem

luxas:
Hi @junaid18183 and @jiangpengcheng :wave:!
Thanks for trying Ignite out, and for the PR :smile:
It's true, we haven't tested it on RHEL, so this needs to get fixed.

[...]

PR #131, opened 2019-07-11
Added the -F flag in mkfs.ext4 command. 

junaid18183:
To force mke2fs to create a filesystem, even if the specified device is not a partition on a block special device, or if other parameters do not make sense


junaid18183:
fixes - https://github.com/weaveworks/ignite/issues/130 

```
# /home/ec2-user/ignite image import weaveworks/ignite-ubuntu:latest
INFO[0000] Created image with ID "d7225bf8913affe4" and name "weaveworks/ignite-ubuntu:latest"
```

PR #132, opened 2019-07-11
Expose firecracker sockets

patrobinson:
This exposes the Firecracker socket files (API, Metrics and Logs) to the host by mapping them into the (already mounted) `/var/lib/firecracker/vm/$id/firecracker` directory.

This is useful for our use case because we want to use the Firecracker metadata service to expose AWS Credentials. These credentials need to be refreshed so cannot be set statically.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -29,7 +29,7 @@ const (
 	IGNITE_TIMEOUT = 10
 
 	// In-container path for the firecracker socket
-	SOCKET_PATH = "/tmp/firecracker.sock"
+	FIRECRACKER_SOCKET_PATH = "firecracker"
 
 	// In-container path for the firecracker log FIFO
 	LOG_FIFO = "/tmp/firecracker_log.fifo"
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -28,14 +28,13 @@ const (
 	// Additional timeout in seconds for docker to wait for ignite to save and quit
 	IGNITE_TIMEOUT = 10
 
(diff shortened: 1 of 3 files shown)


pkg/constants/constants.go
```diff
@@ -29,13 +29,14 @@ const (
 	IGNITE_TIMEOUT = 10
 
 	// In-container path for the firecracker socket
-	SOCKET_PATH = "/tmp/firecracker.sock"
+	FIRECRACKER_SOCKET_PATH = "firecracker"
```
luxas: Can we please avoid creating an extra directory for now to keep the code a bit simpler?
Please attach these sockets at the `/var/lib/firecracker/vm/$id/firecracker.sock` etc. path directly

pkg/metadata/metadata.go
```diff
@@ -97,6 +97,10 @@ func processUID(obj meta.Object, c *client.Client) error {
 	if err := os.MkdirAll(dir, constants.DATA_DIR_PERM); err != nil {
 		return fmt.Errorf("failed to create directory for ID %q: %v", uid, err)
 	}
+	
+	if err := os.MkdirAll(paths.Join(dir, FIRECRACKER_SOCKET_PATH), constants.DATA_DIR_PERM); err != nil {
```
luxas: this can be removed with the above design

pkg/constants/constants.go
patrobinson: Sure thing

patrobinson:
I realised I'm not cleaning up `firecrackerSocketPath` which may result in a VM failing to start again

Issue #135, opened 2019-07-11
ignite run is failing on ssh key gen on FIPS enabled machine


junaid18183:
The failure is happening if the host machine is fips enabled 

```
[root@XXXXXX# ignite run weaveworks/ignite-ubuntu     --cpus 2     --memory 1GB     --ssh     --name my-vm
INFO[0000] Docker image "weaveworks/ignite-kernel:4.19.47" not found locally, pulling...
INFO[0003] Imported OCI image "weaveworks/ignite-kernel:4.19.47" (49.0 MB) to kernel image with UID "a81e3c5b1130b712"
INFO[0004] Removed VM with name "my-vm" and ID "b8fd329361ee5432"
command ["ssh-keygen" "-q" "-t" "ed25519" "-N" "" "-f" "/var/lib/firecracker/vm/b8fd329361ee5432/id_b8fd329361ee5432"] exited with "ED25519 keys are not allowed in FIPS mode\r\n": exit status 255
```

luxas:
Thanks for the report. Commented on the PR, we can target this for v0.4.1

luxas:
This is fixed with #136, and will be available in v0.4.1. Thanks :tada:!

PR #136, opened 2019-07-11
Changing the ssh key gen algorithm for FIPS machines

junaid18183:
Changing the ssh key gen algorithm for FIPS machines 
Fixes the - https://github.com/weaveworks/ignite/issues/135

### DIFF
--- a/pkg/metadata/vmmd/fips.go
+++ b/pkg/metadata/vmmd/fips.go
@@ -1,0 +1,18 @@
+package vmmd
+
+import (
+	"os"
+)
+
+// FIPSEnabled returns true if running in FIPS mode.
+// currently it just checks the system wide /etc/system-fips file present or not.
+// We can improve it later.
+func FIPSEnabled() bool {
+	fips_file := "/etc/system-fips"
+	if _, err := os.Stat(fips_file); err != nil {
+		if os.IsNotExist(err) {
+			return false
+		}
+	}
+	return true
+}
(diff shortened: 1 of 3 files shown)


pkg/metadata/vmmd/fips.go
```diff
@@ -0,0 +1,18 @@
+package vmmd
+
+import (
+	"os"
+)
+
+// FIPSEnabled returns true if running in FIPS mode.
+// currently it just checks the system wide /etc/system-fips file present or not.
+// We can improve it later.
+func FIPSEnabled() bool {
```
luxas: Please move this to `pkg/util`, that seems like a better place for it.

pkg/metadata/vmmd/vm.go
```diff
@@ -193,9 +193,13 @@ func (md *VM) ClearIPAddresses() {
 // Generate a new SSH keypair for the vm
 func (md *VM) newSSHKeypair() (string, error) {
 	privKeyPath := path.Join(md.ObjectPath(), fmt.Sprintf(constants.VM_SSH_KEY_TEMPLATE, md.GetUID()))
-
 	// Use ED25519 instead of RSA for performance (it's equally secure, but a lot faster to generate/authenticate)
-	_, err := util.ExecuteCommand("ssh-keygen", "-q", "-t", "ed25519", "-N", "", "-f", privKeyPath)
+	ssh_key_algorithm := "ed25519"
```
luxas: Please add a note here like this:

```go
// TODO: In future versions, let the user specify what key algorithm to use through the API types
```

pkg/metadata/vmmd/fips.go
junaid18183: Fixed

pkg/metadata/vmmd/vm.go
junaid18183: Fixed

junaid18183:
Thanks for the review. Have implemented the changes.

Issue #85, opened 2019-07-01
Use github.com/vishvananda/netlink instead of execing to `ip`


luxas:
https://github.com/vishvananda/netlink is the de-facto Go package to use for manipulating network interfaces.

luxas:
The relevant code which should be converted to use https://github.com/vishvananda/netlink instead of execing to `ip` is here: https://github.com/weaveworks/ignite/blob/master/pkg/container/network.go

Help from the community here would be super-appreciated!

luxas:
Moving this out of v0.5.0; we won't have time for this before releasing v0.5.0.

PR #146, opened 2019-07-12
Don't fail when there are unsupported net interfaces in the container

### LINKED ISSUE #133: network setup failed: timed out waiting for the condition
Just tried following the startup guide, but it errors out without any further information. Debug logs show nothing helpful. I'm on gentoo, current kernel 5.1.16, kvm + docker setup and running, so it shouldn't be a missing config option or something like that.

```$ sudo ignite run weaveworks/ignite-ubuntu --name test --cpus 2 --memory 1GB --size 6GB --ssh --interactive --log-level=debug
DEBU[0000] Ensuring image weaveworks/ignite-ubuntu:latest exists, or importing it...
DEBU[0000] cache: counted 0 image objects
DEBU[0000] Found image with UID 23c5a4699dfac464
DEBU[0000] Ensuring kernel weaveworks/ignite-kernel:4.19.47 exists, or importing it...
DEBU[0000] cache: counted 0 kernel objects
DEBU[0000] Found kernel with UID 79028168da762017
DEBU[0000] cache: counted 0 VM objects
DEBU[0000] cache: counted 0 VM objects
DEBU[0000] cache: counted 0 image objects
DEBU[0000] cache: counted 0 kernel objects
DEBU[0000] Client.Set; UID: "bc841e6e58a42f77", Kind: VM
DEBU[0000] cache: storing VM object
DEBU[0000] GenericStorage.Set: Key "/vm/bc841e6e58a42f77" did not exist before, set the created field
INFO[0001] Created VM with ID "bc841e6e58a42f77" and name "test"
INFO[0002] Started Firecracker VM "bc841e6e58a42f77" in a container with ID "f6547ff385023505922e07e84dbe81f67c8fba643a7089cac0d07538405df65e"
bc841e6e58a42f77
error: network setup failed: timed out waiting for the condition```

### PR DESCRIPTION
**Pull request by luxas**
Fixes #133 
(I believe)

The problem was that when ignite tried to parse the networking for the following container...

```
$ docker run -it busybox ip addr
...
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
2: sit0@NONE: <NOARP> mtu 1480 qdisc noop qlen 1000
    link/sit 0.0.0.0 brd 0.0.0.0
8: eth0@if9: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.2/16 brd 172.17.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet 192.168.252.10 peer 192.168.252.9/32 scope global tun0
       valid_lft forever preferred_lft forever
    inet6 fe80::5eae:1d4:2b4b:ee4/64 scope link stable-privacy
       valid_lft forever preferred_lft forever
```

...it tried to find the address of the first interface `sit0@NONE`, but that failed (obviously). The problem was that the `networkSetup` function did not try the other interfaces, just retry and do the same thing after one more second.

This PR fixes so that ignite tries all network interfaces.

/cc @twelho @danielcb

### DIFF
--- a/pkg/container/firecracker.go
+++ b/pkg/container/firecracker.go
@@ -36,9 +36,9 @@ func ExecuteFirecracker(md *vmmd.VM, dhcpIfaces []DHCPInterface) error {
 		cmdLine = constants.VM_DEFAULT_KERNEL_ARGS
 	}
 
-	firecrackerSocketPath := path.Join(md.ObjectPath(), FIRECRACKER_API_SOCKET)
-	logSocketPath := path.Join(md.ObjectPath(), LOG_FIFO)
-	metricsSocketPath := path.Join(md.ObjectPath(), METRICS_FIFO)
+	firecrackerSocketPath := path.Join(md.ObjectPath(), constants.FIRECRACKER_API_SOCKET)
+	logSocketPath := path.Join(md.ObjectPath(), constants.LOG_FIFO)
+	metricsSocketPath := path.Join(md.ObjectPath(), constants.METRICS_FIFO)
 	cfg := firecracker.Config{
 		SocketPath:      firecrackerSocketPath,
 		KernelImagePath: path.Join(constants.KERNEL_DIR, md.GetKernelUID().String(), constants.KERNEL_FILE),
--- a/pkg/container/network.go
+++ b/pkg/container/network.go
@@ -47,6 +47,7 @@ func SetupContainerNetworking() ([]DHCPInterface, error) {
 		}
 		if retry {
 			// we got an error, but let's ignore it and try again
+			log.Infof("Got an error while trying to set up networking, but retrying: %v", err)
 			return false, nil
 		}
 		// the error was fatal, return it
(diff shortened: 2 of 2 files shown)


twelho:
This will work fine for the first "pass" of the interfaces, where it checks every one, skips invalid ones and adds valid ones, but if a valid interface is not ready yet, it will be skipped if some other valid interface happens to be ready (as `interfacesCount` becomes > 0).

I suggest if encountering an invalid interface, you add it to `ignoreInterfaces` and set the retry flag to retry the loop again, until a pass is encountered where no interfaces got added to ignored.

You should also check the boolean in `takeAddress`, which signals if an address can be retrieved from the interface. If not, that calls for a retry (without ignoring the interface). It's not perfect however, as that will get stuck if the interface has no address.

Was there an issue where `ignite-spawn` was fast enough to catch an interface before Docker gave it its address? Otherwise we could just use the no addresses flag to ignore the interface.

The correct way would be to retry the interfaces that are down, giving them time to go up first. All interfaces which are up and have an invalid address configuration should be ignored.

luxas:
> This will work fine for the first "pass" of the interfaces, where it checks every one, skips invalid ones and adds valid ones, but if a valid interface is not ready yet, it will be skipped if some other valid interface happens to be ready (as interfacesCount becomes > 0).

This means we would support multiple interfaces. IMO we don't at the moment. We should create an issue towards that end, and fix edge cases like this in all the places that might be relevant. Right now it's on a "best-effort" basis. Without this PR, _one_ interface support is blocked (which is worse), hence I suggest merging this PR now, and following up on multiple interfaces once we're thought through that correctly and implemented tests to ensure it stays supported.

> Was there an issue where ignite-spawn was fast enough to catch an interface before Docker gave it its address?

Yes.

twelho:
OK, it works for now, but we need to come up with a better solution later to actually support multiple interfaces. I would still suggest checking interfaces without _any_ addresses (the `takeAddress` boolean) and retrying them, as `ignite-spawn` might be fast enough to not detect any addresses for any interfaces without it, leaving the VM again without connection. If the interface has an address, but it's incorrect/invalid, just ignore it (by continuing and adding to `ignoreInterfaces`). Should be easy enough to implement.

twelho:
We'll worry about ongoing interface configs/proper multi-interface support later.

luxas:
This all will be better with #85 fixed, too

luxas:
> We'll worry about ongoing interface configs/proper multi-interface support later.

Please create an issue with your thoughts about it and tag v0.5.0

Issue #181, opened 2019-07-15
Require the kernel version to be numeric


luxas:
Right now, if the `strings` binary is unavailable, somehow the kernel version field ends up becoming `/bin/bash: strings: command not found` or similar. We should match the output against a regexp or similar to avoid bad data like that.

FWIW, I just added earlier today to the installation instructions to install `binutils` to avoid the root cause.

PR #189, opened 2019-07-16
Add `-o pipefail` to bash when checking kernel version

twelho:
This enables bash to return an error if any of the commands in the
pipeline fail (such as when the `strings` binary is missing).
Fixes https://github.com/weaveworks/ignite/issues/181.


twelho:
FYI `pipefail` also works with `ksh` and `zsh` in addition to `bash`, so this solution is quite universal. Requiring the kernel version to be numeric or match a regexp is not a good solution in my opinion, as custom kernels can have multiple string tags in the version etc.

PR #212, opened 2019-07-17
duplicate bracket

silenceshell:
One level bracket is enough for this.


twelho:
Hi @silenceshell :wave:!
Good catch, thanks!
LGTM :+1:

PR #215, opened 2019-07-18
Create internal API types, and use them

luxas:
This PR:
 - copies the existing external v1alpha1 API types into `pkg/apis/ignite`, where the external types should be
 - creates conversions between `v1alpha1` and internal
 - removes some small, old hacks we had to workaround the absence of internal types
 - makes the application use internal types everywhere
 - makes the client and storage always return internal types
 - adds Serializer.DefaultInternal to allow defaulting internal types as before

This allows us to roll out more supported external types, e.g. v1alpha2, without disrupting v1alpha1 support. Also, it will allow us to streamline validation, helper method registration, and defaulting.

For more information about how this works, see my talk from KubeCon Shanghai, along with the minimal sample-config implementation:
 - https://speakerdeck.com/luxas/configuring-your-kubernetes-cluster-on-the-next-level
 - https://youtu.be/klHBzISZkCw
 - https://github.com/luxas/sample-config


luxas:
We'll see, not at the moment. We'll for sure make multiple external types as the meta package matures.

PR #225, opened 2019-07-19
Make the Client/Storage able to perform strategic merge patches

### LINKED ISSUE #186: Add `.Patch` to Storage/Client
Like `k8s.io/client-go` has it

 - ~~`.Post` should add a net-new object, and return an error if it already exists~~
 - ~~`.Put` overwrites an existing object for the specified UID~~
 - `.Patch` only updates one or a set of specified fields/values, using a JSON strategic merge patch

This should be done for the `pkg/storage` rework in v0.5.0

luxas:
Based on the functions in `k8s.io/apimachinery` :)

Fixes #186


hack/cobra.go
```diff
@@ -7,15 +7,16 @@ import (
 
 	"github.com/spf13/cobra/doc"
 	"github.com/weaveworks/ignite/cmd/ignite/cmd"
+	"github.com/weaveworks/ignite/pkg/providers"
 )
 
 func main() {
+	providers.Populate()
```
twelho: This needs an error check

pkg/storage/cache.go
```diff
@@ -87,6 +87,13 @@ func (c *cache) Set(gvk schema.GroupVersionKind, obj meta.Object) error {
 	return c.storage.Set(gvk, obj)
 }
 
+// Patch performs a strategic merge patch on the object with the given UID, using the byte-encoded patch given
+func (s *cache) Patch(gvk schema.GroupVersionKind, uid meta.UID, patch []byte) error {
+	// TODO: Should we do something here to cache the change? I don't think so, but...
+	// Just passthrough to the storage here
+	return s.storage.Patch(gvk, uid, patch)
```
twelho: The cache should probably re-cache the object after it has been modified via a patch, but we can implement that in the cache rework together with invalidation.

luxas: yes, exactly

luxas:
Fixed the comment, merging

Issue #119, opened 2019-07-10
Checksum/timestamp based cache invalidation


twelho:
Currently the cache invalidates only on a differing file count. To support detecting configuration changes, the cache needs to track file checksums/timestamps for invalidation.

PR #227, opened 2019-07-19
Support checksum-based Cache invalidation, improve cache's object handling

twelho:
With this the cache supports automatic invalidation on access, `RawStorage` uses the modification timestamp to detect changes. Fixes https://github.com/weaveworks/ignite/issues/119.


twelho:
@luxas is out of office, so as this is ready, going ahead and merging.

Issue #142, opened 2019-07-12
support configuring listen address and protocol


BenTheElder:
It seems the `--port` flag hard codes `0.0.0.0` (on the host side?) and doesn't seem to reference TCP/UDP/..., docker's equivilant flag supports a format like `127.0.0.1:80:80/TCP`, I'd love to be able to at least specify the bind address on the host side.

https://github.com/weaveworks/ignite/blob/45de648fb62850d230d56256a2047bfffac65bab/pkg/apis/meta/v1alpha1/net.go#L19

luxas:
@BenTheElder good idea. We could initially add a `BindAddress net.IP` (and maybe later, optionally `Protocol string`)
How urgent request is this? Should we get it into a `v0.4.x` patch release to unblock kind?
If so, I'm happy to add it. Or maybe you want to send the PR :wink:

BenTheElder:
I don't think we're blocked by it, but I prefer to only bind the API server(s) to localhost by default since they may be unpatched and not particularly locked down :sweat_smile: 

What we did is just ~match the CRI spec: https://github.com/kubernetes/cri-api/blob/0b0c4484b3d3cc29ce8890ba0de35d571db071a2/pkg/apis/runtime/v1alpha2/api.pb.go#L392

https://github.com/kubernetes-sigs/kind/blob/9205bac4a8377fb7d301a8c14650592b85a34f58/pkg/container/cri/types.go#L58

I'd be happy to PR this when I get a chance to work on this further, possibly this weekend :sweat_smile:

PR #232, opened 2019-07-21
implement `ignite exec`

### LINKED ISSUE #65: Consider adding `ignite exec [vm] [command]`
It would:
a) use ssh internally
b) require a command, unlike `ssh`

Don't know if we want to do this or not

BenTheElder:
This is a pretty simple / MVP implementation, however it is enough to perform basic exec (over ssh) against a running ignite VM.

TODO:
- inherit out/err/in writers from the cobra command (other commands seem to be ignoring these inputs anyhow...)
- docs (?)
- support for setting environment variables (also we should consider other `docker exec` options? though `-i` already conflicts ...)
- more extensive testing
- consider porting `ignite ssh` to use the same packages

How this works:
- We open an ssh connection to the VM and create a session with golang.org/x/crypto/ssh
- We then request a pseudo tty in the session and wire up stdin / stdout / stderr to the host
- Since ssh requires sending the command as a single string, `args[1:]` are converted to `args[1] + strings.Join(shellEscape(args[2:...], " ")` and run that way (pseudo code)

Additionally, the exit code from the command is preserved if possible.

```console
$ bin/ignite exec centos echo '"hello world"'
"hello world"

$ bin/ignite exec centos -- /bin/bash -c 'echo hello world "$HOSTNAME"'
hello world 50a7f9aa26409ab3
```

Fixes #65


BenTheElder:
tracking issue: https://github.com/weaveworks/ignite/issues/142

alexellis:
This looks very useful Ben 👍 Out of interest, is it required for KinD to work?

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, fmt.Errorf("unable to read private key: %v", err)
+	}
+
+	// Create the Signer for this private key.
+	return ssh.ParsePrivateKey(key)
+}
+
+func newSSHConfig(publicKey ssh.Signer, timeout uint32) *ssh.ClientConfig {
+	return &ssh.ClientConfig{
+		User: "root",
```
alexellis: Is there a scenario where `root` may not be the user wanted for ssh - VMs with "disable root login"?

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+		Do not wrap this error so the caller can check for the exit code
+		If the remote server does not send an exit status, an error of type *ExitMissingError is returned.
+		If the command completes unsuccessfully or is interrupted by a signal, the error is of type *ExitError.
+		Other error types may be returned for I/O problems.
+	*/
+	return session.Run(joinShellCommand(command))
+}
+
+// joinShellCommand joins command parts into a single string safe for passing to sh -c (or SSH)
+func joinShellCommand(command []string) string {
```
alexellis: @rgee0 @matipan is this method/package any use to us with faas-cli or ofc-bootstrap?

cmd/ignite/run/exec.go
BenTheElder: I think so, but for the moment this matches `ignite ssh` which also assumes this.

BenTheElder:
> This looks very useful Ben 👍 Out of interest, is it required for KinD to work?

kind requires some way to execute commands, we could implement this in kind some other way (driving SSH? Implementing this logic there?), but I think it's more generally useful to have this in ignite.

---

One other note vs `docker exec`, `docker exec` will treat everything after the container name/ID as part of the command. IE `docker exec centos bash --version` works with `--version` as a argument to bash, not docker vs this PR requiring `ignite exec centos -- bash --version`. 

We could implement this but I think we'd need our own flag processing step, I don't think pflag / cobra alone can support this

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+	joined := command[0]
+	if len(command) == 1 {
+		return joined
+	}
+	for _, arg := range command[1:] {
+		// NOTE: we need to escape nested single quotes
```
BenTheElder: Slightly stale comment from before using the shellescape package.

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+	"io/ioutil"
+	"net"
+	"os"
+	"path"
+	"time"
+
+	api "github.com/weaveworks/ignite/pkg/apis/ignite"
+	"github.com/weaveworks/ignite/pkg/constants"
+	"github.com/weaveworks/ignite/pkg/util"
+
```
twelho: Remove this blank line and run `make tidy` again :slightly_smiling_face:

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+	vm      *api.VM
+	command []string
+}
+
+func (ef *ExecFlags) NewExecOptions(vmMatch string, command ...string) (eo *execOptions, err error) {
+	eo = &execOptions{
+		ExecFlags: ef,
+		command:   command,
+	}
+	eo.vm, err = getVMForMatch(vmMatch)
```
twelho: There could be a blank line between the closing brace and `eo.vm` to help readability

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// If an external identity file is specified, use it instead of the internal one
+	privKeyFile := eo.IdentityFile
+	if len(privKeyFile) == 0 {
+		privKeyFile = path.Join(eo.vm.ObjectPath(), fmt.Sprintf(constants.VM_SSH_KEY_TEMPLATE, eo.vm.GetUID()))
+		if !util.FileExists(privKeyFile) {
+			return fmt.Errorf("no private key found for VM %q", eo.vm.GetUID())
+		}
+	}
+	signer, err := newSignerForKey(privKeyFile)
```
twelho: Here a blank line as well

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	// Create an SSH client, and connect, we will use this to exec
+	config := newSSHConfig(signer, eo.Timeout)
+	client, err := ssh.Dial("tcp", net.JoinHostPort(ipAddrs[0].String(), "22"), config)
+	if err != nil {
+		return fmt.Errorf("failed to dial: %v", err)
+	}
+
+	// run the command, DO NOT wrap this error as the caller can check for the command exit
```
twelho: run -> Run

cmd/ignite/run/exec.go
twelho: The disable root login option default in most distributions prevents password-based authentication, but SSH keys should work just fine. We should also open an issue about allowing user selection in `ssh` and `exec`.

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	defer session.Close()
+
+	// get a pty
+	// TODO: should these be based on the host terminal?
+	// TODO: should we request something other than xterm?
+	// TODO: we should probably configure the terminal modes
+	modes := ssh.TerminalModes{}
+	if err := session.RequestPty("xterm", 80, 40, modes); err != nil {
+		return fmt.Errorf("request for psuedo terminal failed: %v", err)
```
twelho: psuedo -> pseudo

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+		return fmt.Errorf("failed to create session: %v", err)
+	}
+	defer session.Close()
+
+	// get a pty
+	// TODO: should these be based on the host terminal?
+	// TODO: should we request something other than xterm?
+	// TODO: we should probably configure the terminal modes
+	modes := ssh.TerminalModes{}
+	if err := session.RequestPty("xterm", 80, 40, modes); err != nil {
```
twelho: Defaulting to `xterm` is fine for now, we could later also allow changing this via a flag. Anyways this is non-blocking for this initial implementation.

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+	// connect input / output
+	// TODO: these should come from the cobra command instead of hardcoding os.Stderr etc.
+	stderr, err := session.StderrPipe()
+	if err != nil {
+		return fmt.Errorf("failed to connect stderr: %v", err)
+	}
+	go io.Copy(os.Stderr, stderr)
+	stdout, err := session.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("failed to connect stderr: %v", err)
```
twelho: stderr -> stdout

cmd/ignite/run/exec.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	go io.Copy(os.Stdout, stdout)
+	stdin, err := session.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("failed to connect stderr: %v", err)
```
twelho: stderr -> stdin

cmd/ignite/run/exec.go
twelho: You can go ahead and remove the comment then :+1:

cmd/ignite/cmd/vmcmd/exec.go
```diff
@@ -0,0 +1,46 @@
...
+			}())
+		},
+	}
+
+	addExecFlags(cmd.Flags(), ef)
+	return cmd
+}
+
+func addExecFlags(fs *pflag.FlagSet, ef *run.ExecFlags) {
+	fs.StringVarP(&ef.IdentityFile, "identity", "i", "", "Override the vm's default identity file")
```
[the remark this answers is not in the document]
BenTheElder: this matches the ssh command's flags exactly at the moment, should we update it there too?

cmd/ignite/run/exec.go
BenTheElder: done

cmd/ignite/run/exec.go
BenTheElder: done

cmd/ignite/run/exec.go
BenTheElder: done

cmd/ignite/run/exec.go
BenTheElder: fixed

cmd/ignite/run/exec.go
BenTheElder: fixed

cmd/ignite/run/exec.go
BenTheElder: updated to a less stale comment

BenTheElder:
rebased

luxas:
We can follow-up the possible improvements here in an other issue later, but for now, merging this to get us started. Next, integrate with kind :wink:

PR #234, opened 2019-07-23
New Storage implementation for the Manifest Directory and inter-Storage Synchronization

twelho:
This PR adds the initial pieces for the new manifest directory support. It just adds new code alongside the existing path, switching over to this new implementation should be done in a separate PR.
- [x] `SyncStorage`
- [x] `WatchStorage`
- [x] `inotify` watchers
- [x] `ManifestRawStorage`
- [ ] `ManifestStorage`
- [ ] `gitops` reworked
- [ ] documentation
...

We might split these into multiple PRs if this one becomes too bloated. I also want to write some documentation with illustrations describing how this all works.


twelho:
The new Storage workflow is now complete with the `SyncStorage`, `WatchStorage`, `inotify` and `ManifestRawStorage` and it _works_! :tada:

There are a couple of bugs/missing features I'll be fixing tomorrow, and GitOps is temporarily disabled until I get around to reworking it to support this (probably also tomorrow). We also need a proper `ManifestStorage` implementation to make this easy to operate, 4792926 currently has a test setup for watching `/etc/firecracker/manifests` recursively for changes and propagating them to `/var/lib/firecracker`, but more testing (and separate unit tests) are still needed before I consider this stable. With over 3,400 added lines this could definitely be separated into smaller PRs before merging.

twelho:
The current `watcher` implementation is based on https://github.com/fsnotify/fsnotify, which doesn't support the `IN_CLOSE_WRITE` operand and uses `IN_MODIFY` instead. This causes a flood of events during `WatchStorage` writes (`Set` operation`), very likely resulting in an infinite write loop.

Will switch `watcher` over to https://github.com/rjeczalik/notify and use `IN_CLOSE_WRITE`, which should only send one event when the file being written to is closed.

cmd/ignite/cmd/hang.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+	"github.com/spf13/cobra"
+	"github.com/weaveworks/ignite/pkg/providers"
+	"io"
+	"os"
+	"os/signal"
+	"sync"
+)
+
+func NewCmdHang(out io.Writer) *cobra.Command {
```
luxas: remove this before merge?

pkg/providers/storage.go
```diff
@@ -5,14 +5,36 @@ import (
...
 	"github.com/weaveworks/ignite/pkg/storage"
 	"github.com/weaveworks/ignite/pkg/storage/cache"
+	"github.com/weaveworks/ignite/pkg/storage/manifest"
+	"github.com/weaveworks/ignite/pkg/storage/sync"
+	"github.com/weaveworks/ignite/pkg/storage/watch"
 )
 
 // Storage is the default storage implementation
 var Storage storage.Storage
+var SS *sync.SyncStorage
```
luxas: rename this to something better, SS doesn't say me anything

pkg/storage/sync/storage.go
```diff
@@ -0,0 +1,148 @@
...
+
+// SyncStorage is a Storage implementation taking in multiple Storages and
+// keeping them in sync. Any write operation executed on the SyncStorage
+// is propagated to all of the Storages it manages (including the embedded
+// one). For any retrieval or generation operation, the embedded Storage
+// will be used (it is treated as read-write). As all other Storages only
+// receive write operations, they can be thought of as write-only.
+type SyncStorage struct {
+	storage.Storage
+	storages    []storage.Storage
```
luxas: rename to writeOnlyStorages or similar?

cmd/ignite/cmd/hang.go
luxas: or, maybe better, rename it to `daemon` for now

pkg/providers/storage.go
```diff
@@ -5,14 +5,36 @@ import (
...
 
 func SetCachedStorage() error {
 	Storage = cache.NewCache(
 		storage.NewGenericStorage(
 			storage.NewDefaultRawStorage(constants.DATA_DIR), scheme.Serializer))
 	return nil
 }
+
+// TODO: Special constructor for this setup
+func SetTestManifestStorage() error {
```
luxas: remove `Test` from here, and call this when running `ignite daemon`

pkg/storage/format.go
```diff
@@ -0,0 +1,14 @@
+package storage
+
+type Format uint8
```
luxas: TODO: godoc comments

pkg/storage/manifest/raw/rawstorage.go
```diff
@@ -0,0 +1,152 @@
...
+	"os"
+	"path/filepath"
+	"strconv"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/weaveworks/ignite/pkg/storage"
+	"github.com/weaveworks/ignite/pkg/util"
+)
+
+type MappedRawStorage interface {
```
luxas: godoc comment

pkg/storage/manifest/raw/rawstorage.go
```diff
@@ -0,0 +1,152 @@
...
+	storage.RawStorage
+
+	AddMapping(key storage.Key, path string)
+	GetMapping(path string) (storage.Key, error)
+	RemoveMapping(key storage.Key)
+}
+
+func NewManifestRawStorage(dir string) MappedRawStorage {
```
luxas: for clarity, let the struct implementation be called `mappedRawStorage` and the func be `NewMappedRawStorage` then

luxas: in other words, go all-in on the Mapped word instead of manifest here

pkg/storage/manifest/storage.go
```diff
@@ -1,25 +1,19 @@
...
-// ManifestStorage implements the storage interface for GitOps purposes
-type ManifestStorage struct {
-	storage.Storage
-	gitRaw *ManifestRawStorage
-}
-
-func (s *ManifestStorage) Sync() (UpdatedFiles, error) {
-	return s.gitRaw.Sync()
-}
+//func NewManifestStorage(dataDir string) *ManifestStorage {
```
luxas: use /* */ instead to avoid that many line changes

pkg/storage/sync/storage.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/weaveworks/ignite/pkg/client"
+	"github.com/weaveworks/ignite/pkg/storage"
+	"github.com/weaveworks/ignite/pkg/storage/watch"
+	"github.com/weaveworks/ignite/pkg/storage/watch/update"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type UpdateStream chan update.Update
+
+const updateBuffer = 4096 // How many updates to buffer
```
luxas: document why you chose this value

pkg/storage/sync/storage.go
```diff
@@ -0,0 +1,165 @@
...
+			errC <- result{i, f(s)}
+		}(i, s) // NOTE: This requires i and s as arguments, otherwise they will be evaluated for one Storage only
+	}
+
+	for i := 0; i < len(ss.storages); i++ {
+		if result := <-errC; result.error != nil {
+			if err == nil {
+				err = fmt.Errorf("SyncStorage: error in Storage %d: %v", result.int, result.error)
+			} else {
+				err = fmt.Errorf("%v\n%29s %d: %v", err, "and error in Storage", result.int, result.error)
```
luxas: is %29s desired here?

pkg/storage/sync/storage.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return
+}
+
+func (ss *SyncStorage) monitorFunc() {
+	defer log.Debug("SyncStorage: monitoring thread stopped")
+
+	// This is the internal client for propagating updates
+	c := client.NewClient(ss)
```
luxas: is this run only once, or multiple times?  if once, then cache the client in the syncstorage object

pkg/storage/sync/storage.go
```diff
@@ -0,0 +1,165 @@
...
+	// This is difficult to do though, as we have don't know which state is the latest
+	// For now, only update the state on write when Ignite is running
+	// TODO: Trigger a modify event for all existing files on startup?
+	for {
+		if upd, ok := <-ss.eventStream; ok {
+			switch upd.Event {
+			case update.EventModify, update.EventCreate:
+				// First load the Object using the Storage given in the update,
+				// then set it using the client constructed above
+				if obj, err := client.NewClient(upd.Storage).Dynamic(upd.APIType.GetKind()).Get(upd.APIType.GetUID()); err != nil {
```
luxas: this code is logically correct, but hard to read. factor out `obj, err := ...` into its own line, and then do the `:= c.Set` command on its own dedicated line, too. the `continue` is there so nothing would change when doing so

pkg/storage/sync/storage.go
```diff
@@ -0,0 +1,165 @@
...
+				if err := c.Dynamic(upd.APIType.GetKind()).Delete(upd.APIType.GetUID()); err != nil {
+					log.Errorf("Failed to delete Object with UID %q: %v", upd.APIType.GetUID(), err)
+					continue
+				}
+			}
+
+			// Send the update to the listeners
+			ss.updateStream <- upd.Update
+		} else {
+			return
```
luxas: log something here? what happens if the update wasn't ok? ignite crashes? deadlocks?

pkg/storage/watch/monitor.go
```diff
@@ -0,0 +1,27 @@
+package watch
+
+import "sync"
+
+type Monitor struct {
```
luxas: godoc here
does this have to be in this package or could it be under e.g. 'util/monitor`? it doesn't seem watch-related really

pkg/storage/watch/storage.go
```diff
@@ -0,0 +1,176 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/yaml"
+)
+
+// Storage is an interface for persisting and retrieving API objects to/from a backend
+// One Storage instance handles all different Kinds of Objects
+type WatchStorage interface {
+	// WatchStorage extends the Storage interface
+	storage.Storage
+	// GetTrigger returns a hook that can be used to detect a watch event
```
luxas: ?

pkg/storage/watch/storage.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/weaveworks/ignite/pkg/storage/manifest/raw"
+	"github.com/weaveworks/ignite/pkg/storage/watch/update"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/yaml"
+)
+
+// Storage is an interface for persisting and retrieving API objects to/from a backend
+// One Storage instance handles all different Kinds of Objects
+type WatchStorage interface {
```
luxas: godoc here this is misleading, please update

pkg/storage/watch/update/associated.go
```diff
@@ -0,0 +1,10 @@
+package update
+
+import (
+	"github.com/weaveworks/ignite/pkg/storage"
+)
+
+type AssociatedUpdate struct {
```
luxas: godoc

pkg/storage/watch/update/event.go
```diff
@@ -0,0 +1,25 @@
...
+	case 1:
+		return "CREATE"
+	case 2:
+		return "DELETE"
+	}
+
+	return "MODIFY"
+}
+
+type FileUpdate struct {
```
luxas: godoc

pkg/storage/watch/update/update.go
```diff
@@ -0,0 +1,10 @@
+package update
+
+import (
+	meta "github.com/weaveworks/ignite/pkg/apis/meta/v1alpha1"
+)
+
+type Update struct {
```
luxas: godoc

pkg/storage/watch/watch.go
```diff
@@ -0,0 +1,196 @@
...
+	notify.InCreate:     update.EventCreate,
+	notify.InDelete:     update.EventDelete,
+	notify.InCloseWrite: update.EventModify,
+}
+
+type eventStream chan notify.EventInfo
+type UpdateStream chan *update.FileUpdate
+type watches []string
+
+type watcher struct {
```
luxas: directoryWatcher or inotifyWatcher would be a better name here IMO, I didn't understand what `watcher` did before looking at it

pkg/storage/watch/watch.go
```diff
@@ -0,0 +1,196 @@
+package watch
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/rjeczalik/notify"
```
luxas: any specific reason you chose this package?
it seems like it doesn't support recursive watching, right?

pkg/storage/watch/watch.go
```diff
@@ -0,0 +1,196 @@
...
+// addition to the generated watcher, it can be used to populate
+// MappedRawStorage fileMappings
+func newWatcher(dir string) (w *watcher, files []string, err error) {
+	w = &watcher{
+		dir:     dir,
+		events:  make(eventStream, eventBuffer),
+		updates: make(UpdateStream, eventBuffer),
+	}
+
+	if err = w.start(&files); err != nil {
```
luxas: I think returning `files, err := w.start()` is way clearer instead of the pointer pattern

luxas:
please drop the dependencies.svg update from this PR, we don't need that now.
let's get this ready to merge by today evening

twelho:
Dropped the dependency graph update.

cmd/ignite/cmd/hang.go
twelho: I'll rename it to `daemon`

pkg/providers/storage.go
twelho: It's `SyncStorage` now, added comments to it as well

pkg/storage/sync/storage.go
twelho: The `storages` slice contains _all_ storages, so both the write-only ones as well as the read-write one (which is embedded as well). Basically every storage that the writes get propagated to.

twelho:
```golang
storages:     append(wStorages, rwStorage),
```

pkg/providers/storage.go
twelho: Done

pkg/storage/format.go
twelho: Done

pkg/storage/manifest/raw/rawstorage.go
twelho: :+1: it's now called `GenericMappedRawStorage` matching all our other (Raw)Storages implementing their interfaces.

pkg/storage/manifest/raw/rawstorage.go
twelho: Done

pkg/storage/manifest/storage.go
twelho: Done

pkg/storage/sync/storage.go
twelho: I just chose it by feeling :sweat_smile: but I'll add something

pkg/storage/sync/storage.go
twelho: Yes, it indents the errors correctly for readability.

pkg/storage/sync/storage.go
twelho: It's run once, in a goroutine. Caching it in SyncStorage brings no benefits, as it's only used in monitorFunc and doesn't get removed.

pkg/storage/sync/storage.go
twelho: Done, it's a little easier to read when the client generation and object fetching are on separate lines.

pkg/storage/sync/storage.go
twelho: Added logging, Ignite will now issue a warning and discard updates if the channel becomes full. If the update failed somehow, Ignite doesn't send it into the channel at all and issues an error from the code above this sender. The goroutines never crash or deadlock.

pkg/storage/watch/monitor.go
twelho: Yes I was thinking about moving it there, good catch!

pkg/storage/watch/storage.go
twelho: Yes, the initial plan of having SyncStorage talk directly to the RawStorages didn't pan out very well, so I took the liberty of implementing WatchStorage. It sits as the Storage implementation between SyncStorage and MappedRawStorage and manages the watcher (`inotify`), the MappedStorage mappings and triggering SyncStorage to update via itself.

pkg/storage/watch/storage.go
twelho: Updated in https://github.com/weaveworks/ignite/pull/242.

pkg/storage/watch/update/associated.go
twelho: Added in https://github.com/weaveworks/ignite/pull/242.

pkg/storage/watch/update/event.go
twelho: Done

pkg/storage/watch/update/update.go
twelho: Done in https://github.com/weaveworks/ignite/pull/242.

pkg/storage/watch/watch.go
twelho: It's only used internally (it's local), so I like to keep the name short, but I'll add a godoc comment to explain what it is and does.

pkg/storage/watch/watch.go
twelho: `w.start(nil)` is used inside the goroutine as well, by using the pointer I avoid performing a potentially expensive filename validation during `inotify` monitoring. This is why a pointer is used:
```golang
if files != nil {
	// Only include files with a valid suffix
	if validSuffix(info.Name()) {
		*files = append(*files, path)
	}
}
```

pkg/storage/watch/watch.go
twelho: `inotify` itself doesn't support recursive watches. For this reason there are no good golang packages supporting that feature.

The few small ones that might support it don't provide low-level enough access to `inotify` to monitor `IN_CLOSE_WRITE`, which is the ideal way to fire a single event after the whole write is complete instead of a flood of events for every small write operation. https://github.com/rjeczalik/notify doesn't support recursive watching, but provides the `IN_CLOSE_WRITE` I need.

`watcher` _does_ however support recursive watching, it will dynamically add and remove watches for all subdirectories:
```golang
if fi.IsDir() {
	if err := w.addWatch(event.Path()); err != nil {
```

PR #241, opened 2019-07-29
Implement read-write GitOps

twelho:
This PR contains a collection of improvements to `SyncStorage` and GitOps support for it.
- [x] Event batching using `batcher`
- [x] GitOps read-write support (committing and pushing)
- [x] Send MODIFY events for all existing files when starting Ignite
- [x] `watcher` rework for `batcher`


twelho:
:+1:

PR #245, opened 2019-07-30
Extract watcher, batcher and monitor into pkg/util

luxas:
This makes them reusable in a generic manner, and not tied to `pkg/storage`


pkg/storage/watch/update/update.go
```diff
@@ -2,11 +2,21 @@ package update
...
 	meta "github.com/weaveworks/ignite/pkg/apis/meta/v1alpha1"
+	"github.com/weaveworks/ignite/pkg/storage"
+	"github.com/weaveworks/ignite/pkg/util/watcher"
 )
 
 // Update bundles an Event with an
 // APIType for Storage retrieval.
 type Update struct {
-	Event   Event
+	Event   watcher.Event
```
twelho: Event though `watcher` uses the Events the most, they are not `watcher` specific, the same events are used to detect changes for Objects in `WatchStorage` and `SyncStorage`. Should we move `Event` to `watcher` or keep it in `update`, to be generic?

luxas: I think it's better to keep the types fairly closely to where they are used. I did not understand the purpose of the types when seeing them grouped in the `update` package. I think they are a good fit in this package, at least until something more significant changes in the codebase.

twelho: As it seems to work for `WatchStorage` and `SyncStorage` in terms of dependencies, LGTM :+1:. I'll go ahead and merge.

luxas:
Rebased; merging now

Issue #207, opened 2019-07-17
Create a website with documentation


luxas:
Using Netlify, MkDocs or ReadTheDocs for example.
I like how the https://eksctl.io site is created, and would be very happy to do the same thing for Ignite.

dholbach:
https://github.com/weaveworks/flux/pull/2152 could serve as an example.

luxas:
Reopening this until we have a public read-the-docs endpoint (and/or a custom domain).
@dholbach can you make this happen?

dholbach:
https://ignite.readthedocs.io/en/latest/ is live!

luxas:
Thank you!! Then we're done here :smile:

dholbach:
Can this issue be closed and we track all upcoming bits in new issues?

dholbach:
Ah yes. :-D

PR #246, opened 2019-07-30
Start off read-the-docs effort for documentation

dholbach:
Re-use some of the work for flux, targetting #207.
Fixes: https://github.com/weaveworks/ignite/issues/207

TODO:

- [x] include `api/` - currently not indexed, as it's not inside `docs/` (linking to GH right now)
- [ ] register ignite on readthedocs and import
- [x] fix videos (Videos (e.g. in "How to use Ignite to run VMs") have their link visible next to the thumbnail, it might be good to just have the thumbnail as the chapters above each video explain them)
- [x] fix bare URLs in changelog docs


dholbach:
If you want to test it, just pull it and run `make serve-docs` and point your browser to `http://localhost:8000/_build/html/index.html`.

dholbach:
Screenshot from 2019-07-31 16-44-46

twelho:
This is defintely the right direction, it looks good! :tada:
A couple of things I noticed when running the docs locally:
- Videos (e.g. in "How to use Ignite to run VMs") have their link visible next to the thumbnail, it might be good to just have the thumbnail as the chapters above each video explain them
- Half of the items in the "CLI reference" section have their short descriptions next to the command and the other half don't. IMO it's clearer if we just have the commands without descriptions in the list, but is it possible to make subheadings for the different subcommands (such as VM/Kernel/Image commands)?
- Some of the changelogs don't render their links correctly, might also be a mistake on our side

In the CLI refrence, is this (50% of commands with descriptions) the result of `make tidy`, or have you modified them by hand?

dholbach:
> This is defintely the right direction, it looks good! tada
> A couple of things I noticed when running the docs locally:
> 
>     * Videos (e.g. in "How to use Ignite to run VMs") have their link visible next to the thumbnail, it might be good to just have the thumbnail as the chapters above each video explain them

Ok, will add as a TODO item.

 
>     * Half of the items in the "CLI reference" section have their short descriptions next to the command and the other half don't. IMO it's clearer if we just have the commands without descriptions in the list, but is it possible to make subheadings for the different subcommands (such as VM/Kernel/Image commands)?

I used the descriptions that are currently listed under https://github.com/weaveworks/ignite/blob/master/docs/cli/ignite.md#see-also

As you can see in https://github.com/weaveworks/ignite/pull/246/files#diff-87b56d02ca5719fe568e2c07fffda67aR74 the titles will now we automatically generated from the document headings. I used a glob (`*`) to pull in the ones that currently are not listed (that was just an idea).


>     * Some of the changelogs don't render their links correctly, might also be a mistake on our side
> 
> 
> In the CLI refrence, is this (50% of commands with descriptions) the result of `make tidy`, or have you modified them by hand?

Well spotted. Looks like "bare URLs" are not working - adding to TODO as well.

dholbach:
@luxas @twelho With the rest I'm going to need help.

Issue #63, opened 2019-06-10
Logging everywhere


twelho:
- [x] `errutils` removed

luxas:
Part of this is done, part of this is for later

luxas:
Now that we have structured logging in v0.4.0, there are a couple of TODOs still:

 - Start using logrus everywhere; add log statements for different levels all over the applications
 - Get rid of `errutils.Check`
 - Pass loglevel through from `ignite` to `ignite-spawn` and Firecracker itself.

PR #247, opened 2019-07-30
Make Ignite use `logrus` everywhere, support setting log level for ignite-spawn and Firecracker

twelho:
Fixes https://github.com/weaveworks/ignite/issues/63.
- [x] Use `logrus` everywhere for logging (`log.Print` -> `log.Info`)
- [x] Refactor `errutils` away
- [x] Remove duplicate `cobra` command errors
- [x] Pass log level through from `ignite` to `ignite-spawn` and Firecracker itself


cmd/ignite-spawn/root.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/spf13/pflag"
+	"github.com/weaveworks/ignite/cmd/ignite/cmd/cmdutil"
+	"github.com/weaveworks/ignite/pkg/logs"
+)
+
+var logLevel = logrus.InfoLevel
+
+// NewIgniteSpawnCommand returns the root command for ignite-spawn
+func NewIgniteSpawnCommand(in io.Reader, out, err io.Writer) *cobra.Command {
+	root := &cobra.Command{
```
luxas: I did not expect to see/use cobra in `ignite-spawn` to keep the resource usage of ignite-spawn down to a minimum.
Please use "plain" pflag instead.

pkg/container/firecracker.go
```diff
@@ -43,6 +45,18 @@ func ExecuteFirecracker(vm *api.VM, dhcpIfaces []DHCPInterface) error {
 		return err
 	}
 
+	// Convert the logrus error level to a Firecracker compatible error level.
+	// Firecracker accepts "Error", "Warning", "Info", and "Debug", case-sensitive.
```
luxas: :+1:

cmd/ignite-spawn/root.go
twelho: Refactored in 0568c54.

PR #250, opened 2019-07-31
Fork v1alpha1 API to become v1alpha2

twelho:
Also include `v1alpha2` in the generators run via `make autogen`.

Fixes: #251


luxas:
Thanks!

PR #259, opened 2019-07-31
Activate v1alpha2 in the scheme

luxas:
@twelho noted that this also needs to be done to activate the new types (ref: #250)


luxas:
To get an idea of the size of this: it increases the binary size with 0.2MB (by including v1alpha2 in the builds)

PR #275, opened 2019-08-05
Support external volumes (block devices) in Ignite VMs

twelho:
This PR adds support for exposing host block devices inside Ignite VMs. It also sports internal automounting support via `/etc/fstab`. The volumes are now declared in the API as `.spec.storage.volumes` and `.spec.storage.volumeMounts` k8s style, so this changes `v1alpha2`.
Fixes https://github.com/weaveworks/ignite/issues/76.

cc @luxas

PR #281, opened 2019-08-06
`make tidy` after implementing volume support

twelho:
This includes all changes by `make tidy` on top of https://github.com/weaveworks/ignite/pull/275.


luxas:
I fixed autogen/tidy on latest master; and rebased #275, so this is not needed anymore.

PR #282, opened 2019-08-06
Generate API docs for multiple versions

luxas:
cc @dholbach @twelho


luxas:
The build is green! I'm merging this now to keep going, this is very mechanical.

PR #279, opened 2019-08-06
refactor: use netlink instead of exec'ing

alexeldeib:
Fixes https://github.com/weaveworks/ignite/issues/85

PR #285, opened 2019-08-07
Clean up ignite-spawn networking, don't depend on `errors` package

twelho:
This extends @alexeldeib's amazing work in https://github.com/weaveworks/ignite/pull/279.
I noticed we still depend on the `errors` package and have a potential nil indexing in `ignite-spawn`'s networking, so went ahead and cleaned up a little.

cc @luxas


twelho:
Yes, `iproute2` is not required in the `weaveworks/ignite` image anymore. Removed in 5af057c.

PR #286, opened 2019-08-07
Two-way SyncStorage support for ManifestStorage

twelho:
This enables changes to propagate both ways between `/var/lib/firecracker` and `/etc/firecracker/manifests` :tada:

There are a couple of issues that need to be fixed for reliability:
- [ ] `ignite-spawn` IP updates cause events, which for some reason makes Ignite attempt VM startup again
- [ ] Deletion is also performed twice, we probably should limit object updates to one Storage

cc @luxas


luxas:
Closing this, as we went an other way, with `ignite(d)` doing the patch with the IP addresses/etc.

PR #299, opened 2019-08-08
Support configuring the bind address and protocol for port binds

twelho:
Fixes https://github.com/weaveworks/ignite/issues/142. This enables configuring the the listening address on the host and the protocol used when specifying a port binding from both declaratively and from the CLI.

This PR is part of the API changes for v0.5.0.
TODO:
- [ ] Duplicate VM port checks
- [ ] Verify this works in practice

cc @luxas


pkg/apis/meta/v1alpha1/net.go
```diff
@@ -9,49 +10,95 @@ import (
 
 // PortMapping defines a port mapping between the VM and the host
 type PortMapping struct {
-	HostPort uint64 `json:"hostPort"`
-	VMPort   uint64 `json:"vmPort"`
+	BindAddress net.IP   `json:"bindAddress,omitempty"`
```
luxas: this is optional; I'd prefer for this to be a pointer (although, technically maybe omitempty works here)

pkg/runtime/docker/client.go
```diff
@@ -105,8 +106,20 @@ func (dc *dockerClient) AttachContainer(container string) (err error) {
...
-		portBindings[nat.Port(fmt.Sprintf(portFormat, portMapping.VMPort))] = []nat.PortBinding{
+		var hostIP string
+		if portMapping.BindAddress != nil {
+			hostIP = portMapping.BindAddress.String()
+		}
+
+		protocol := portMapping.Protocol
+		if len(protocol) == 0 {
+			// Docker uses TCP by default
+			protocol = meta.ProtocolTCP
```
luxas: create a `const DefaultProtocol = ProtocolTCP` in meta?

pkg/apis/meta/v1alpha1/net.go
```diff
@@ -9,49 +10,95 @@ import (
...
+	HostPort    uint64   `json:"hostPort"`
+	VMPort      uint64   `json:"vmPort"`
+	Protocol    Protocol `json:"protocol,omitempty"`
```
luxas: I'm think that omitempty doesn't work with custom marshallers, but idk. We can test. If so, then we need to use a pointer here.

pkg/apis/meta/v1alpha1/net.go
```diff
@@ -9,49 +10,95 @@ import (
...
 }
 
 // PortMappings represents a list of port mappings
 type PortMappings []PortMapping
 
 var _ fmt.Stringer = PortMappings{}
 
+var errInvalidPortMappingFormat = fmt.Errorf("port mappings must be of form [<bind address>:]<host port>:<VM port>[/<protocol>]")
+
 func ParsePortMappings(input []string) (PortMappings, error) {
```
luxas: I'm strongly of the opinion that we should not duplicate code here that we already have vendored.
Please use https://github.com/docker/go-connections/blob/master/nat/nat.go#L126 for this task; and turn `map[Port][]PortBinding` from there into `meta.PortMappings`.

pkg/apis/meta/v1alpha1/net.go
```diff
@@ -74,6 +121,44 @@ func (p PortMappings) String() string {
...
+	}
+
+	return "", fmt.Errorf("invalid protocol: %q", input)
+}
+
+func (p Protocol) String() string {
+	return string(p)
+}
+
+func (p Protocol) MarshalJSON() ([]byte, error) {
```
luxas: this is not needed; that's the default

pkg/runtime/docker/client.go
```diff
@@ -105,8 +106,20 @@ func (dc *dockerClient) AttachContainer(container string) (err error) {
 func (dc *dockerClient) RunContainer(image string, config *runtime.ContainerConfig, name string) (string, error) {
 	portBindings := make(nat.PortMap)
```
luxas: factor this out to a `portBindingsToPortMap` function please?

pkg/apis/meta/v1alpha1/net.go
twelho: `net.IP` is a `[]byte`, which can be `nil`, so omitempty works just fine.

pkg/runtime/docker/client.go
twelho: The default protocol is runtime implementation specific, for Docker it just happens to be TCP, so I wouldn't make it generic.

luxas: ok, sure

pkg/apis/meta/v1alpha1/net.go
twelho: Removed in a182160.

pkg/runtime/docker/client.go
twelho: Factored in a182160.

pkg/apis/meta/v1alpha1/net.go
twelho: Done in a182160.

pkg/apis/meta/v1alpha1/net.go
twelho: Now that we're using the standard mashaler and Docker's parser always adding the proto, this isn't an issue.

twelho:
This doesn't include validation for reading the port mappings declaratively, is that a blocker for merging? I'll add a TODO to implement it anyway as we've not had it until now either.

twelho:
> Can you open a new issue to track missing validation functions?
Sure thing: https://github.com/weaveworks/ignite/issues/301

PR #300, opened 2019-08-08
Runtime state set by Ignite (StartVM), ignite-spawn only handles shutdown

twelho:
With this `ignite-spawn` no longer writes to the VM metadata file on startup, it only performs cleanup on after VM shutdown. The running state, container ID, start time and VM IP addresses are now set by `StartVM` after starting the container (and waiting for `ignite-spawn`).

cc @luxas

### DIFF
--- a/cmd/ignite-spawn/ignite-spawn.go
+++ b/cmd/ignite-spawn/ignite-spawn.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"net"
 	"os"
 	"path"
 
@@ -28,11 +27,13 @@ func decodeVM(vmID string) (*api.VM, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	vm, ok := obj.(*api.VM)
 	if !ok {
 		return nil, fmt.Errorf("object couldn't be converted to VM")
 	}
-	// Explicitely set the GVK on this object
+
+	// Explicitly set the GVK on this object
 	vm.SetGroupVersionKind(api.SchemeGroupVersion.WithKind(api.KindVM.Title()))
 	return vm, nil
 }
@@ -47,7 +48,7 @@ func StartVM(vm *api.VM) error {
 	// Serve DHCP requests for those interfaces
 	// This function returns the available IP addresses that are being
 	// served over DHCP now
-	ipAddrs, err := container.StartDHCPServers(vm, dhcpIfaces)
+	_, err = container.StartDHCPServers(vm, dhcpIfaces)
 	if err != nil {
 		return err
 	}
@@ -55,11 +56,6 @@ func StartVM(vm *api.VM) error {
 	// Serve metrics over an unix socket in the VM's own directory
 	metricsSocket := path.Join(vm.ObjectPath(), constants.PROMETHEUS_SOCKET)
 	serveMetrics(metricsSocket)
-
-	// Update the VM status and IP address information
-	if err := patchRunning(vm, ipAddrs); err != nil {
-		return fmt.Errorf("failed to patch VM state: %v", err)
-	}
 
 	// Patches the VM object to set state to stopped, and clear IP addresses
 	defer patchStopped(vm)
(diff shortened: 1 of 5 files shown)


twelho:
This has an issue: `ignite-spawn` only manages to clear the state about 30% of the time. An unsuccessful clear leaves Ignite in a limbo, where the VM cannot be started or stopped anymore without editing the state by hand.

cmd/ignite-spawn/ignite-spawn.go
```diff
@@ -47,7 +48,7 @@ func StartVM(vm *api.VM) error {
 	// Serve DHCP requests for those interfaces
 	// This function returns the available IP addresses that are being
 	// served over DHCP now
-	ipAddrs, err := container.StartDHCPServers(vm, dhcpIfaces)
+	_, err = container.StartDHCPServers(vm, dhcpIfaces)
```
luxas: Please cleanup the code that returns this unnecessary slice

cmd/ignite-spawn/ignite-spawn.go
```diff
@@ -56,11 +57,6 @@ func StartVM(vm *api.VM) error {
 	metricsSocket := path.Join(vm.ObjectPath(), constants.PROMETHEUS_SOCKET)
 	serveMetrics(metricsSocket)
```
luxas: Maybe use a waitgroup or something here to actually unblock the goroutine later when we know the firecracker process has started?

pkg/operations/start.go
```diff
@@ -132,12 +132,19 @@ func StartVM(vm *api.VM, debug bool) error {
 		return err
 	}
 
-	// TODO: Enable this when ignite-spawn doesn't write to the file anymore
-	//if err := providers.Client.VMs().Set(vm); err != nil {
-	//	return err
-	//}
+	result, err := providers.Runtime.InspectContainer(containerID)
```
luxas: You might want to add a comment above that we need to do this differently for CNI; there the IP address should be got from the CNI netns. We can follow that up

cmd/ignite-spawn/ignite-spawn.go
```diff
@@ -80,22 +76,14 @@ func StartVM(vm *api.VM) error {
...
-func patchRunning(vm *api.VM, ipAddrs []net.IP) error {
-	return patchVM(vm, func(patchVM *api.VM) error {
-		patchVM.Status.Running = true
-		patchVM.Status.IPAddresses = ipAddrs
-		return nil
-	})
-}
-
 func patchStopped(vm *api.VM) error {
 	return patchVM(vm, func(patchVM *api.VM) error {
```
luxas: Now that the patch is only done in one place, and all the desired values are predictable, you can apply a static patch that you've defined in a string

cmd/ignite-spawn/ignite-spawn.go
twelho: We can only get close to the startup place, as the Firecracker call is blocking. `ignite-spawn` cannot actually track Firecracker's state (without listening to some sockets), so this cannot be done at the moment.

pkg/operations/start.go
twelho: I'll add that

cmd/ignite-spawn/ignite-spawn.go
twelho: Done in 872516a.

pkg/operations/start.go
twelho: Added in 94f7149.

cmd/ignite-spawn/ignite-spawn.go
twelho: Done in 872516a.

twelho:
The issue with `ignite-spawn`'s state clearing is solved with 872516a, the patcher was using metadata from when `ignite-spawn` started up as its base, which had been read while under update by Ignite's `StartVM` resulting in a race condition. That has now been prevented by using a static patch.

cmd/ignite-spawn/ignite-spawn.go
luxas: Ok, good for now

twelho:
> LGTM, thanks! Please create issues with the TODO followups from this PR

Done: https://github.com/weaveworks/ignite/issues/302 and https://github.com/weaveworks/ignite/issues/303

Issue #77, opened 2019-06-18
Add support for the metadata service in `ignite run`


luxas:
Firecracker has a [metadata service](https://github.com/firecracker-microvm/firecracker/blob/master/docs/mmds.md) that allows for passing arbitrary JSON data to the VM, where it client-side can be fetched from a hardcoded `169.254.169.254` IP endpoint.

We should allow the user to specify this metadata at runtime.

patrobinson:
Wouldn't it be better to expose `firecracker.sock` say at `/var/lib/firecracker/vm/$id/firecracker.sock`?
This way you could dynamically update the metadata and access other features of the Firecracker API.

luxas:
Yes, that is also a good idea (approved #132 :smile:), but at VM start time (IIRC), you need to set whether the MMDS should be enabled or not, and by default we have it disabled for less resource usage.
So we'd most probably like to expose some high-level knob to set the data, too

patrobinson:
@luxas oh cool, I'll work on that next

luxas:
Thank you @patrobinson :sparkles:! Let us know if you need any help.

luxas:
Hi @patrobinson :wave: any progress here? Do you want help on getting this up and running?

patrobinson:
I had a look at this but it wasn't trivial. It's not currently a focus for me sorry.

luxas:
Ok, no problem. Would you like some smaller work item to help out on @patrobinson instead or are you too busy at the moment?

patrobinson:
Thanks @luxas for being so forthcoming, but at this stage I'm not looking to actively contribute. That may change in the future and I'll let you know.

luxas:
Gotcha. No problem :smile:!

alexeldeib:
@luxas what are you thinking here? user passes in json on vm creation and that gets applied with a PUT after VM is up, or introducing an ignite CLI command to pass through arbitrary data to MMDS? Or something else?

alexeldeib:
I started on this 😄 Would appreciate guidance though

luxas:
Thanks! I'll comment on the PR :+1:

PR #304, opened 2019-08-08
[WIP] feat: expose mmds

alexeldeib:
ref https://github.com/weaveworks/ignite/issues/77

I took an initial stab at exposing MMDS. I'm not super familiar with vsock -- does this approach make sense? is a constant okay for CID/path? I can pull those into the constants file.

One issue with MMDS is it requires a firecracker binary compiled with vsock support. This is fairly easy on a local machine but I am still struggling to build it in the docker image. So definitely don't merge this 😉 

Additionally, since the metadata must be injected to firecracker VM from inside the container, I've opted to pass the metadata as Environment variable at Boot time (should hook in properly to start as well as run).

### DIFF
--- a/cmd/ignite/cmd/vmcmd/create.go
+++ b/cmd/ignite/cmd/vmcmd/create.go
@@ -70,6 +70,7 @@ func addCreateFlags(fs *pflag.FlagSet, cf *run.CreateFlags) {
 	// Register flags for simple types (int, string, etc.)
 	fs.Uint64Var(&cf.VM.Spec.CPUs, "cpus", cf.VM.Spec.CPUs, "VM vCPU count, 1 or even numbers between 1 and 32")
 	fs.StringVar(&cf.VM.Spec.Kernel.CmdLine, "kernel-args", cf.VM.Spec.Kernel.CmdLine, "Set the command line for the kernel")
+	fs.StringVar(&cf.VM.Spec.Metadata, "metadata", cf.VM.Spec.Metadata, "Enable MMDS and use this JSON value for creation metadata")
 
 	// Register more complex flags with their own flag types
 	cmdutil.SizeVar(fs, &cf.VM.Spec.Memory, "memory", "Amount of RAM to allocate for the VM")
--- a/pkg/apis/ignite/types.go
+++ b/pkg/apis/ignite/types.go
@@ -185,7 +185,8 @@ type VMSpec struct {
 	// Specifying a path in SSH.Generate means "use this public key"
 	// If SSH.PublicKey is set, this struct will marshal as a string using that path
 	// If SSH.Generate is set, this struct will marshal as a bool => true
-	SSH *SSH `json:"ssh,omitempty"`
+	SSH      *SSH   `json:"ssh,omitempty"`
+	Metadata string `json:"metadata,omitempty"`
 }
 
 type VMImageSpec struct {
--- a/pkg/container/firecracker.go
+++ b/pkg/container/firecracker.go
@@ -22,12 +22,15 @@ import (
 // ExecuteFirecracker executes the firecracker process using the Go SDK
 func ExecuteFirecracker(vm *api.VM, dhcpIfaces []DHCPInterface) error {
 	drivePath := vm.SnapshotDev()
+	mmdsData := os.Getenv("FC_META")
+	enableMmds := mmdsData != ""
 
 	networkInterfaces := make([]firecracker.NetworkInterface, 0, len(dhcpIfaces))
 	for _, dhcpIface := range dhcpIfaces {
 		networkInterfaces = append(networkInterfaces, firecracker.NetworkInterface{
 			MacAddress:  dhcpIface.MACFilter,
 			HostDevName: dhcpIface.VMTAP,
+			AllowMMDS:   enableMmds,
 		})
 	}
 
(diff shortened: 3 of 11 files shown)


pkg/container/firecracker.go
```diff
@@ -119,19 +128,23 @@ func ExecuteFirecracker(vm *api.VM, dhcpIfaces []DHCPInterface) error {
...
-	//}
+	}
 
 	if err := m.Start(ctx); err != nil {
 		return fmt.Errorf("failed to start machine: %v", err)
 	}
-	defer m.StopVMM()
 
+	if enableMmds {
+		//cfg.KernelArgs = fmt.Sprintf("%s -device vhost-vsock-pci,id=vhost-vsock-pci0,guest-cid=3")
```
alexeldeib: cleanup

pkg/container/firecracker.go
```diff
@@ -22,12 +22,15 @@ import (
 // ExecuteFirecracker executes the firecracker process using the Go SDK
 func ExecuteFirecracker(vm *api.VM, dhcpIfaces []DHCPInterface) error {
 	drivePath := vm.SnapshotDev()
+	mmdsData := os.Getenv("FC_META")
```
alexeldeib: cleanup; take from vm spec

alexeldeib:
Hm! I’ve already a few things, notably that CID should be unique per VM. Not sure how to track that though, maybe some sort of global cache? Or check existence first somehow?

luxas:
Thanks for this contribution @alexeldeib ✨💯🎉!
I will take a look at this PR on Monday and give feedback :). I don't think this will make v0.5.0 that we're in the process pf pushing out at the moment, but we hope to release v0.6.0 shortly afterwards :)

alexeldeib:
(notes to self as I work through this)

I think I should also use kernel args `-device vhost-vsock-pci,id=vhost-vsock-pci0,guest-cid=$CID` instead of mounting `/dev/vsock` like i've done?

alexeldeib:
(had some issues with my dev environment, will work on finishing this up soon)

cmd/ignite/cmd/vmcmd/create.go
```diff
@@ -70,6 +70,7 @@ func addCreateFlags(fs *pflag.FlagSet, cf *run.CreateFlags) {
 	// Register flags for simple types (int, string, etc.)
 	fs.Uint64Var(&cf.VM.Spec.CPUs, "cpus", cf.VM.Spec.CPUs, "VM vCPU count, 1 or even numbers between 1 and 32")
 	fs.StringVar(&cf.VM.Spec.Kernel.CmdLine, "kernel-args", cf.VM.Spec.Kernel.CmdLine, "Set the command line for the kernel")
+	fs.StringVar(&cf.VM.Spec.Metadata, "metadata", cf.VM.Spec.Metadata, "Enable MMDS and use this JSON value for creation metadata")
```
luxas: We'd also want to point this to a file; and use that content instead of requiring the JSON to be inline.

pkg/apis/ignite/types.go
```diff
@@ -185,7 +185,8 @@ type VMSpec struct {
 	// Specifying a path in SSH.Generate means "use this public key"
 	// If SSH.PublicKey is set, this struct will marshal as a string using that path
 	// If SSH.Generate is set, this struct will marshal as a bool => true
-	SSH *SSH `json:"ssh,omitempty"`
+	SSH      *SSH   `json:"ssh,omitempty"`
+	Metadata string `json:"metadata,omitempty"`
```
luxas: would `[]byte` be a better format?
also add this field to v1alpha2 please

pkg/runtime/docker/client.go
```diff
@@ -127,12 +127,12 @@ func (dc *dockerClient) RunContainer(image string, config *runtime.ContainerConf
 	}
 
 	stopTimeout := int(config.StopTimeout)
-
 	c, err := dc.client.ContainerCreate(context.Background(), &container.Config{
 		Hostname:    config.Hostname,
 		Tty:         true, // --tty
 		OpenStdin:   true, // --interactive
 		Cmd:         config.Cmd,
+		Env:         config.Env,
```
luxas: Cleanup later?

pkg/operations/start.go
```diff
@@ -74,6 +75,8 @@ func StartVM(vm *api.VM, debug bool) error {
 			runtime.BindBoth("/dev/net/tun"),        // Needed for creating TAP adapters
 			runtime.BindBoth("/dev/kvm"),            // Pass through virtualization support
 			runtime.BindBoth(vm.SnapshotDev()),      // The block device to boot from
+			runtime.BindBoth("/dev/vhost-vsock"),
```
luxas: only if metadata is present?

luxas:
Firecracker v0.18.0 released with "real" vsock support :) https://github.com/firecracker-microvm/firecracker/blob/master/docs/vsock.md

alexeldeib:
🙁 sorry for leaving this dangling.

I haven't had the time to properly invest (or i've been diverting it elsehwere, unfortunately). I'm going to close this. If I am able to brush up on vsock and implement the real thing, I'll open it in a new PR. 

Still appreciate the project and enjoy playing with it 🙂

Issue #306, opened 2019-08-09
`ignite rm -f` fails to remove the container


luxas:
Due to a race condition in the docker container implementation:

```
ignite rm -f f4c3114f94acafe1
FATA[0000] failed to remove container for VM "f4c3114f94acafe1": Error response from daemon: removal of container b78b8fe5e7997852445248b24c213a261ab1becae96725db073b8b4c54d37c05 is already in progress
```

cc @twelho

PR #320, opened 2019-08-12
Fix `ignite rm -f` for a running VM without `--debug`

twelho:
The docker client has no way to perform container existence detection and removal at the same time, which triggers a potential race condition with the container auto-removal. Avoid the issue by ignoring removal errors for now, added comments and TODO to improve upon this later.
Fixes https://github.com/weaveworks/ignite/issues/306. 

cc @luxas


luxas:
This needs a (small) rebase

Issue #322, opened 2019-08-12
`ignite create --config` does not accept `v1alpha2` specifications


twelho:
Running `ignite create --config conf.yml` where `conf.yml` contains
```yaml
apiVersion: ignite.weave.works/v1alpha2
kind: VM
metadata:
  name: test-vm
  uid: fb733199f239a759
spec:
  image:
    oci: weaveworks/ignite-ubuntu
  cpus: 2
  diskSize: 3GB
  memory: 800MB
```
results in
`FATA[0000] unable to decode ignite.weave.works/v1alpha2, Kind=VM into *ignite.VM`.

Do you have an idea @luxas as to why this happens? `v1alpha2` is registered in the scheme.

PR #323, opened 2019-08-12
Wrap the strict deserializer as a codec

luxas:
Fixes: #322 

The bytes are of the external version (e.g. v1alpha2), and we're trying to deserialize it into the internal version. A _deserializer_ can only deserialize into the right target object you pass it (hence the error about the types mismatching), whereas a _decoder_ (part of a _codec_) figures out what the byte content is, uses the deserializer to unmarshal into that specific struct, and then perform defaulting and conversion into the requested (this time internal) object.

Hence, I'm wrapping the (raw) serializer that's constructed in `CodecForVersions`, which gives back a _decoder_ which automatically does conversion and defaulting as we'd expect.


twelho:
LGTM :+1: 
Merging...

Issue #328, opened 2019-08-12
Use Github code-owners


luxas:
@dholbach pointed out we could use https://help.github.com/en/articles/about-code-owners

PR #331, opened 2019-08-13
add simple CODEOWNERS file

dholbach:
Here Dennis and Lucas are both added for all areas of the code.
Fixes: #328


dholbach:
(It's what we went with for Scope too: https://github.com/weaveworks/scope/blob/master/.github/CODEOWNERS)

Issue #295, opened 2019-08-08
Importing large images doesn't work


luxas:
As reported by @neith00 (IIRC) on Slack.

If you import a large-ish image (e.g. ~900MB), `ignite` fails strangely with:

```
# ignite create node
INFO[0000] Starting image import...                     
FATA[0010] exit status 2
```

It seems to timeout after ten seconds when importing; although the importing process of a larger image requires more time (naturally).

cc @twelho

luxas:
This has been present already in v0.4.x, so it's not a release blocker for v0.5.0. We'd like to fix it in a v0.5.x patch release, though.

alex-leonhardt:
not sure if there was/is a fix in master but I've just tried this (I had built `ignite` from master) and I cannot reproduce this anymore .. the example isn't my image but it's ~912MB so large enough that it should've triggered the issue .. 

```
vagrant@ubuntu-bionic:/vagrant/bin/amd64$ sudo ./ignite image import mbopm/activemq:latest
INFO[0000] Docker image "mbopm/activemq:latest" not found locally, pulling... 
INFO[0092] Starting image import...                     
INFO[0132] Imported OCI image "mbopm/activemq:latest" (912.0 MB) to base image with UID "08b1f641cf98da3a" 
INFO[0132] Created image with ID "08b1f641cf98da3a" and name "mbopm/activemq:latest" 
vagrant@ubuntu-bionic:/vagrant/bin/amd64$ free -m 
              total        used        free      shared  buff/cache   available
Mem:            985         136         243           0         604         697
Swap:             0           0           0
vagrant@ubuntu-bionic:/vagrant/bin/amd64$ sudo ./ignite image import mbopm/activemq
INFO[0000] Created image with ID "08b1f641cf98da3a" and name "mbopm/activemq:latest" 
vagrant@ubuntu-bionic:/vagrant/bin/amd64$ sudo ./ignite image import mbopm/activemq:5.12.1
INFO[0000] Docker image "mbopm/activemq:5.12.1" not found locally, pulling... 
INFO[0015] Starting image import...                     
INFO[0057] Imported OCI image "mbopm/activemq:5.12.1" (912.0 MB) to base image with UID "e2ae14049c239993" 
INFO[0057] Created image with ID "e2ae14049c239993" and name "mbopm/activemq:5.12.1" 
```

[...]

PR #335, opened 2019-08-14
Automatically optimize the size of an imported image

twelho:
This enables importing large images by oversizing the sparse file block device, importing the contents and then shrinking it to the smallest size possible. Tested with an OCI image with two 10 GB layers.
Fixes https://github.com/weaveworks/ignite/issues/295.

cc @luxas, @neith00, @alex-leonhardt


twelho:
@luxas is out of office, so going ahead and merging now that CI is green.

PR #336, opened 2019-08-14
Fix ignite-spawn's formatting when performing cleanup on VM metadata

twelho:
`ignite-spawn` directly writes the output of `StrategicMergePatch` to the metadata file, resulting in the formatting/indentation being lost (it's all just one unorganized line). With these changes the `StrategicMergePatch` output is re-encoded via the serializer to preserve the `runtime.Object` structure and indentation on write.

cc @luxas


twelho:
@luxas is out of office, so going ahead and merging now that CI is green.

Issue #327, opened 2019-08-12
GitOps should be file-location agnostic


stealthybox:
@twelho mentioned on the [aug12 developer call](https://docs.google.com/document/d/1fv8_WD6qXfvlIq7Bb5raCGyBvc42dNF-l8uaoZzoUYI) that the current GitOps implementation will destroy and re-create the VM if you move a manifest to a different location in the git repo.

It should be possible to re-organize manifests to wherever file paths you want in the git-repo.
As long as the unique information does not change (names/id's), ignite should be able to keep track of the object.

The internal catalog should key resources by their Group/Version/Kind/name or tuple of other UID's.
That should allow ignite to continue managing the same VM even if the file location of the manifest is moved by the user.

stealthybox:
( If we ever intend to make the directory structure or filenames in the Git-repo significant to the meaning of the manifests these can be intentionally added to the identifying tuple of resources. )

PR #341, opened 2019-08-16
FileWatcher: Support internal moves without re-creating and multiple active moves at once

twelho:
This enables the `FileWatcher` to perform internal moves (e.g. a watched directory to its watched subdirectory) by performing a remap of the `MappedRawStorage` instead of having to delete and re-create the VM. Also enables tracking multiple active moves at once, so `mv *.yml subdir/` now works, and creation and deletion updates for files moved in/out of watched directories now work properly.
Fixes https://github.com/weaveworks/ignite/issues/327.

cc @luxas


twelho:
Approved by @luxas and CI is green, merging :+1:

PR #342, opened 2019-08-16
Use util.Prefixer in place of manual concatenation, use IGNITE_PREFIX in util.Prefixer

twelho:
Just a small cleanup to still include in v0.5.1. Removes manual concatenations in favor of `util.Prefixer` and makes it use `constants.IGNITE_PREFIX`. Builds on top of https://github.com/weaveworks/ignite/pull/341.

cc @luxas


twelho:
It's tested an I can merge it now before v0.5.1, so going ahead and doing that.

twelho:
This won't be included in v0.5.1.

PR #347, opened 2019-08-19
Split packages so we can extract `gitops-toolkit`

luxas:


pkg/storage/manifest/storage.go
```diff
@@ -1,22 +1,40 @@
...
+	return &ManifestStorage{
+		Storage: ss,
+	}, nil
+}
+
+// NewManifestStorage constructs a new storage that watches unstructured manifests in the specified directory,
+// decodable using the given serializer. However, all changes in the manifest directory, are also propagated to
+// the structured data directory that's backed by the default storage implementation. Writes to this storage are
+// propagated to both the manifest directory, and the data directory.
+func NewTwoWayManifestStorage(manifestDir, dataDir string, ser serializer.Serializer) (*ManifestStorage, error) {
```
twelho: `TwoWay` might be a bit confusing as this doesn't propagate data directory changes to the manifest directory (like the name implies). Is `DataManifestStorage` better?

pkg/storage/watch/storage.go
```diff
@@ -42,8 +40,12 @@ func NewGenericWatchStorage(s storage.Storage) (WatchStorage, error) {
 		return nil, err
 	}
 
+	// TODO: Fix this
+	gvs := s.Serializer().Scheme().PreferredVersionAllGroups()
+	groupName := gvs[0].Group
```
twelho: OK for now

pkg/storage/manifest/storage.go
luxas: It does propagate all the manifest that it knows about from before, so I think this is ok for now at least. We can talk about naming later.

pkg/serializer/serializer.go
```diff
@@ -0,0 +1,179 @@
...
+
+// EncodeJSON encodes the specified object for a specific version to pretty JSON bytes
+func (s *serializer) EncodeJSON(obj runtime.Object) ([]byte, error) {
+	return s.encode(obj, runtime.ContentTypeJSON, true)
+}
+
+func (s *serializer) encode(obj runtime.Object, mediaType string, pretty bool) ([]byte, error) {
+	info, ok := runtime.SerializerInfoForMediaType(s.codecs.SupportedMediaTypes(), mediaType)
+	if !ok {
+		return nil, fmt.Errorf("unable to locate encoder -- %q is not a supported media type", mediaType)
```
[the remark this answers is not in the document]
luxas: This was not new code; just the same as before.

luxas:
Merging to get going with the gitops-toolkit extraction

PR #350, opened 2019-08-19
Fix Docker client port mappings by actually exposing them after binding

twelho:
Until now Ignite has just been creating port mappings but not actually put them into effect using the Docker client. Apparently the ports need to be exposed after binding, this PR fixes that. This is also a prerequisite to fixing the `footloose` integration's SSH access.

cc @luxas


twelho:
CI is green, merging :+1:

PR #357, opened 2019-08-20
Add openSUSE images

aojea:
openSUSE has two different versions that release [independently](https://en.opensuse.org/Lifetime):

* openSUSE Tumbleweed is openSUSE's rolling release, which is constantly updated and always at the 'latest release'
* openSUSE Leap is openSUSE's regular release

This PR adds openSUSE images based on the versions because of the [Dockerfile base images naming](https://hub.docker.com/_/opensuse), i.e. Tumbleweed and Leap, not based in the releases, i.e Leap 15.1, 15.2, ... 

If there is more demand I can add a new parameter to the Makefile, but I think that's enough to have some images working with ignite at this moment.

### DIFF
--- a/images/Makefile
+++ b/images/Makefile
@@ -31,10 +31,12 @@ push-all: build-all
 
 build-all:
 	$(MAKE) ${OP} WHAT=amazon-kernel
-	$(MAKE) ${OP} WHAT=amazonlinux   RELEASE=2     IS_LATEST=true
+	$(MAKE) ${OP} WHAT=amazonlinux   RELEASE=2          IS_LATEST=true
 	$(MAKE) ${OP} WHAT=alpine
+	$(MAKE) ${OP} WHAT=opensuse      RELEASE=leap       IS_LATEST=true
+	$(MAKE) ${OP} WHAT=opensuse      RELEASE=tumbleweed
 	$(MAKE) ${OP} WHAT=ubuntu        RELEASE=16.04
-	$(MAKE) ${OP} WHAT=ubuntu        RELEASE=18.04 IS_LATEST=true
+	$(MAKE) ${OP} WHAT=ubuntu        RELEASE=18.04      IS_LATEST=true
 	$(MAKE) ${OP} WHAT=ubuntu        RELEASE=19.04
-	$(MAKE) ${OP} WHAT=centos        RELEASE=7     IS_LATEST=true
+	$(MAKE) ${OP} WHAT=centos        RELEASE=7          IS_LATEST=true
 	$(MAKE) ${OP} WHAT=kubeadm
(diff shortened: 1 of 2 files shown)


images/Makefile
```diff
@@ -31,10 +31,12 @@ push-all: build-all
 
 build-all:
 	$(MAKE) ${OP} WHAT=amazon-kernel
-	$(MAKE) ${OP} WHAT=amazonlinux   RELEASE=2     IS_LATEST=true
+	$(MAKE) ${OP} WHAT=amazonlinux   RELEASE=2          IS_LATEST=true
 	$(MAKE) ${OP} WHAT=alpine
+	$(MAKE) ${OP} WHAT=opensuse      RELEASE=leap
+	$(MAKE) ${OP} WHAT=opensuse      RELEASE=tumbleweed IS_LATEST=true
```
twelho: We use `IS_LATEST` to signal which image gets the `:latest` tag, not necessarily which image is the latest _release_ (see the Ubuntu image). For me it seems more logical to have the stable leap release cycle as the default (`:latest`), unless it's more common to run tumbleweed in OpenSUSE land.

images/opensuse/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+        iputils \
+        openssh \
+        net-tools \
+        systemd-sysvinit \
+        udev \
+        sudo \
+        wget && \
+    zypper clean --all
+
+RUN systemctl enable sshd
```
twelho: IIRC `systemctl` won't work during the image build (as it requires `systemd` to be running as PID 1). To work around this, create the symlink manually here (`systemctl enable sshd` on a host will show the source and target).

images/opensuse/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Default to tumbleweed distro
+ARG RELEASE=tumbleweed
```
twelho: No need to default as this ARG should always be required/specified during a build.

twelho:
cc @luxas

images/opensuse/Dockerfile
aojea: I tested it locally and it's working for me

```
All repositories have been cleaned up.
Removing intermediate container 36876475bd36
 ---> 5ce4c71f7f9b
Step 4/5 : RUN systemctl enable sshd
 ---> Running in a26e0fbb9803
Removing intermediate container a26e0fbb9803
 ---> 896b98b208f7
```
anyway, I can change it if you want

twelho: If it works for both `leap` and `tumbleweed` I'm fine with that, no need to change :+1:

aojea:
> Also, have you verified you can successfully build the image? We don't have CI set up for that just yet sweat_smile

Yeah, I ran it locally and tested the images sshing into the VMs

images/opensuse/Dockerfile
twelho: You could add a comment above that command for brevity, then I think we're ready to merge

twelho:
LGTM, thanks! :+1:

PR #363, opened 2019-08-26
add completion for ignited

silenceshell:
Add completion for ignited.

```
root@pve:~# ignited # press tab twice here
completion  daemon      gitops      version
```


silenceshell:
@luxas `make tidy` done, please have a look

PR #370, opened 2019-08-30
CNI: create the config directory and add Ignite config if not configured

twelho:
Writes a small CNI configuration similar to `docker-bridge` to the CNI config directory if nothing else is configured (the directory doesn't exist/is empty).

cc @luxas


twelho:
CI is green and we need this for v0.6.0, merging.

PR #374, opened 2019-08-30
GitOps: only change the VM state if it differs from the current one

twelho:
Fixes duplicate state changes causing errors when Ignite updates the repository.

cc @luxas


twelho:
CI is green, merging :+1:

PR #376, opened 2019-08-30
Implement cleanup of CNI networks using the default bridge

luxas:
This also cleans up the rules created by the `bridge` plugin in the default `ignite-containerd-bridge` plugin


luxas:
Merging this now to unblock v0.6.0... :+1:

PR #377, opened 2019-08-30
CNI: fix conflist loading, avoid name collision and nil allocation

twelho:
Follow-up on https://github.com/weaveworks/ignite/pull/376, fixes config loading, as the extension needs to be `.conflist` instead of `.conf` now.

cc @luxas


twelho:
CI is green, merging :+1:

Issue #365, opened 2019-08-26
Ignite(-spawn) leaks device mapper devices


twelho:
It seems like the `device mapper` snapshot cleanup has stopped working at some point (again), it might be a specific command (I suspect `rm -f`) that is causing a leak of dm and loop devices. Noticed after using `Footloose` to repeatedly spin up and tear down a cluster of VMs.

twelho:
`rm -f` works properly, the issue is improper/non-working cleanup if `run` or `start` fail during the call e.g. if any of the OCI images could not be pulled.

chanwit:
addressed in #381

chanwit:
Steps to reproduce In my scenario:

Scenario 1:
I got a new binary compiled with different ignite version say, `0.6.0+1-dirty`.
When running `ignite run` it's going to pull `weaveworks/ignite:0.6.0+1-dirty` which does not exist. Then the VM creation failed.

After that `sudo ignite rm -f <vm_id>`

Then `sudo dmsetup ls`, the dm devices still left on disk.

chanwit:
ping @stealthybox could you please close this one. thank you.

chanwit:
This issue could be closed for now. I'll re-open it if things still wrong.

PR #381, opened 2019-08-31
fix possible dm snapshot leaks

chanwit:
There are some defers that won't be called if an error occurred by `ActivateSnapshot`.
It's gonna defer to `DeactivateSnapshot` anyway even if `ActivateSnapshot` return an error.

### DIFF
--- a/pkg/dmlegacy/vm.go
+++ b/pkg/dmlegacy/vm.go
@@ -83,16 +83,17 @@ func AllocateAndPopulateOverlay(vm *api.VM) error {
 }
 
 func copyToOverlay(vm *api.VM) error {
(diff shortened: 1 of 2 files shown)


chanwit:
Also kindly ping @luxas

chanwit:
Also, could we have this for 0.6.x?

chanwit:
kindly ping @stealthybox
I think we need to confirm this change a bit to make sure that it won't break the current behavior.

stealthybox:
I was able to replicate these issues with the dm snapshots mentioned in #365 using a build on master.

This patch appears to fix it.
In general, this is an improvement in code quality.

Misplaced defer statements are a common source of error-handling issues.
Thanks for the thoughtful change :+1:

PR #385, opened 2019-09-01
default cni network does not support multi-node

silenceshell:
If the network is default cni, I think we cannot access vm's IIP from other computers as the IP is like `172.18.0.0/16`.

### DIFF
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -23,7 +23,6 @@ Automatically installed to `/etc/cni/net.d/10-ignite.conflist` unless you have p
 
 **Pros:**
 
-- **Multi-node support**: CNI implementations can often route packets between multiple physical hosts. External computers can access the VM's IP.
 - **Kubernetes-compatible**: You can use the same overlay networks as you use with Kubernetes, and hence get your VMs on the same network as your containers.
 - **Port mapping support**: This mode supports port mappings from the VM to the host
 


chanwit:
@silenceshell I think this is not true for a multi-host CNI like Weave Net.
Weave Net, for example, offers IPs under 10.32.0.0/16 and accessible across nodes.

silenceshell:
yeah, weave is [an-third-party-cni-plugin](https://github.com/weaveworks/ignite/blob/edd7ee643d3a616b922d84f16fe4d5ebc37c190b/docs/networking.md#an-third-party-cni-plugin), and supports multi-node, while the default cni is bridge and not support multi-node.

```json
{
	"cniVersion": "0.4.0",
	"name": "ignite-containerd-bridge",
	"plugins": [
		{
			"type": "bridge",
			"bridge": "cni0",
			"isGateway": true,
			"isDefaultGateway": true,
			"promiscMode": true,
			"ipMasq": true,
			"ipam": {
				"type": "host-local",
				"subnet": "172.18.0.0/16"
			}
		},
		{
			"type": "portmap",
			"capabilities": {
				"portMappings": true
			}
		}
	]
}
```

chanwit:
oh sorry. I haven't read the whole doc. This LGTM then.

PR #389, opened 2019-09-03
specify the minimum version of make

### LINKED ISSUE #383: Not able to run make build-all
Error : Makefile:20: *** missing separator.  Stop

silenceshell:
fix #383

`undefine` (line 20) exists in GNU make 3.82 and higher, however on some os `make` is still 3.81 or even lower, for example macOS 10.14.

We should check make version in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+MIN_MAKE_VERSION = 3.82
+ifneq ($(MIN_MAKE_VERSION), $(firstword $(sort $(MAKE_VERSION) $(MIN_MAKE_VERSION))))
+$(error this project requires make version $(MIN_MAKE_VERSION) or higher)
+endif
+
 SHELL:=/bin/bash
 UID_GID?=$(shell id -u):$(shell id -g)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,3 +1,7 @@
+ifneq (3.82, $(firstword $(sort $(MAKE_VERSION) 3.82)))
```
stealthybox: I'm just noting here that this is a lexical sort ([ref](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html#index-sorting-words))

This would eventually break with Make version 10, which is not a near-term concern.

chanwit: @silenceshell could you try to compute with `bc` or something like this:

https://stackoverflow.com/a/21491594

silenceshell:
@stealthybox @chanwit 
Thank you for your response. I have updated the code, please help to review :)

chanwit:
LGTM

stealthybox:
Thanks for the love on this patch @silenceshell

PR #416, opened 2019-09-11
improve preflight check and add a containerd test case for ignite run

chanwit:
This PR improves pre-flight checks especially for CNI detection.
Also this PR adds an e2e test case for `containerd` and CNI.


stealthybox:
Tried this out earlier -- pre-flights look good.
Test is failing, but we're working on it
LGTM  :+1:

Issue #423, opened 2019-09-11
CNI setup appears more latent than docker-bridge


stealthybox:
Network connectivity does not require sleeps for `docker+docker-bridge` when performing run, then exec.

`docker+cni` only succeeds locally on my laptop with sleep values >= 2 seconds.

Described in #422 with reproducing tests.

chanwit:
Adding DNS 8.8.8.8 and 8.8.4.4 directly thru --dns (a pending PR) eliminated this wait. 
I'm guessing that what's happening on my machine was probably some services trying to query DNS and failed by timeout. 

The workaround is just adding Ignite to wait for the SSH service to start. So we can make sure that the subsequent commands could be able to `ssh`, and `exec` thru the SSH port.

chanwit:
@stealthybox If you're OK, I think we could postpone this one to the next release?
If so, we could cut release by Monday and do integration test with other projects (which is higher priority).

PR #425, opened 2019-09-12
Store e2e command output /w errors + remove variable sleep

stealthybox:
This makes the test output much more readable. Errors are nested with their command args.
It also allows you to read parallel output.
I haven't enabled parallel, because vm run appears to have a bug using dmsetup in parallel.

Successful commands no longer output to stdout, but we could add a gate for this if desired.

I removed the variable sleep because it appears to have less impact on the docker-bridge behavior in #423 than I previously observed.


chanwit:
LGTM

PR #426, opened 2019-09-13
Make getIPChains more precise and less failure-prone

### LINKED ISSUE #393: Getting a CNI removal error when doing ignite vm stop

stealthybox:
move fmt.Sprintf out of loop
access stat.Options through rawStat[9] with hard-coded index
remove break in the event there are multiple rules per containerID
only parse for IPNet when we are working with the proper ignite CNI rules
  ^ avoids https://github.com/coreos/go-iptables/issues/70

Fixes #393


chanwit:
LGTM

PR #421, opened 2019-09-11
[WIP] Stabilize CNI connectivity, dns and latent

chanwit:
This PR fixes a `containerd` + `CNI` behavior that users cannot connect to the internet.
After investigation, I have found that there's not FORWARD rules to help the CNI bridge plugin to complete this. So I add simple rules for the Ignite network CIDR to allow forwarding.

fixes #418 #423 #424

PR #428, opened 2019-09-15
implement --dns flag to allow overriding nameservers

chanwit:
This PR implements a new flag, `--dns`, to allow overriding DNS nameservers in the VM's `/etc/resolv.conf`.

### DIFF
--- a/cmd/ignite/cmd/vmcmd/create.go
+++ b/cmd/ignite/cmd/vmcmd/create.go
@@ -38,6 +38,7 @@ func NewCmdCreate(out io.Writer) *cobra.Command {
 					--name my-vm \
 					--cpus 2 \
 					--ssh \
+					--dns 8.8.8.8 \
 					--memory 2GB \
 					--size 6GB
 		`, constants.DEFAULT_KERNEL_IMAGE)),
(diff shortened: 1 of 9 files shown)


chanwit:
@stealthybox please review this one instead of #421 
Thank you

stealthybox:
I think the flag interface is a good start.
Need a further review on the annotations -- maybe we should have a struct instead.

For the implementation, I'm not sure if DHCP solution is better than using the runtime's `resolv.conf`.
It's certainly nice that since it's done through the `vmDir/manifest`, it does not need to be re-implemented in the runtime -- that's a really good thing.

Removing this from the milestone for now.

bbros-dev:
> I think the flag interface is a good start.

**_Please_** could Ignite turn away from this path of adding API (cli flags) config data in this 'atomic' fashion?
It leads to nightmares such as this [example (from docker)](https://github.com/chef-cookbooks/docker/blob/master/libraries/helpers_service.rb#L187).

If the issue here is that a different `resolv.conf` file is needed, address that directly:
```bash
--resolv string              Specify a path to a resolv.conf to override the default file.
```
**In fact `copyFiles` `--copy-files` should be sufficient here?**
Happy to elaborate if required.

Issue #438, opened 2019-09-19
Consider disabling DNS resolution for the ignite-spawn container


stealthybox:
It's not actually important for any process in the `ignite-spawn` container to have DNS resolution. This actually increases the attack surface of a vm breakout, as many attacks find control servers and even ex-filtrate data using DNS alone.

We should consider making DNS resolution impractical since we have no need for it.
We can load our `dns.ClientConfig` from a different file in the container.
If we do this with a readonly rootfs, it will make attacks more impractical.

Containerd, does not configure `/etc/resolv.conf` by default.

For the docker runtime, ` --dns 0.0.0.0` is one way to effectively disable lookups:
```shell
docker run -it --rm --dns 0.0.0.0 busybox wget google.com
wget: bad address 'google.com'
```

PR #441, opened 2019-09-19
Fix containerd resolv.conf + DHCP behavior

### LINKED ISSUE #437: DHCPv4 DNS incorrect on host /w systemd-resolved using --runtime=containerd
Currently, the containerd runtime indiscriminately mounts `/etc/resolv.conf` into the container.

On an Ubuntu host using `systemd-resolved`, this will break because `/etc/resolv.conf` points to the NS proxy systemd-resolved running @ `127.0.0.53`/`::ffff:127.0.0.53`.

It's important that we mount a working `/etc/resolv.conf` into the runtime container's mount namespace, because `ignite-spawn` will use it to configure DHCP and the nameservers will end up as the vm's primary DNS servers.

Docker implements some nice logic that detects `["127.0.0.53"]` and uses a file that `systemd-resolved` maintains @ `/run/systemd/resolve/resolv.conf`.

Docker also post-processes any resolvconf it loads from disk.
This is to filter loopback addresses since they won't work well in a fresh network namespace.
If no nameservers are left they append Google's public NS as a fallback.

The results of this post-processing are always written to a new file per-container.
This behavior is well documented here https://docs.docker.com/v17.09/engine/userguide/networking/default_network/configure-dns/.

We should consider implementing a subset of this heuristic for our containerd runtime.
It makes sense at the moment to support `nameserver` and `search` lines.
( More intricate DNS options don't all translate well to a DHCP configuration. )

stealthybox:
Fixes #437

Patches:
- Add heuristic resolv.conf builder supporting nameserver+search
- Fix containerd DHCP using resolvconf + Add id to runtime.RunContainer()

Having the user be able to setup custom DNS is great, but it would be better if it first worked well without any tweaking.

This resolv.conf library supports `/etc/resolv.conf`, systemd-resolved, and will even work with nothing at all.
All loopback nameservers are filtered, and in the event that there's still no usable DNS, it falls back to Google DNS servers. (other default suggestions are welcome)

The containerd runtime then uses this lib to populate the vmDir specfic to the ignite vm's id.
A modification to the runtime interface was needed to pass the id so this could be done.

It might be better to try and use the container's snapshot instead of the vmDir to accomplish this.
However, using the vmDir is a generic solution since adding mounts for the runtime container is simple.
We could use this mechanism to more uniformly configure DHCP /w docker while disabling the default resolver for ignite-spawn as mentioned in #438.

I'd also be fine not actually writing any file and just adding fields to the vm manifest for `dhcp.Nameservers` and `dhcp.Search`.
We can still use readDNSConfig() and its accompanying functions to calculate the proper values from the host.

The benefit of using `/etc/resolv.conf` is it's well integrated with docker networks which the vm will benefit from through it's DHCP config.
______________________
example behavior:
```shell
make build-all
~/Repos/ignite-scratch/ignite-clean.sh
~/Repos/ignite-scratch/iptables-clean-cni-ignite.sh

sudo bin/ignite vm run weaveworks/ignite-ubuntu --log-level debug --ssh --name test441
DEBU[0000] Ensuring image weaveworks/ignite-ubuntu:latest exists, or importing it... 
DEBU[0000] Found image with UID 35c47bfb397b922b        
DEBU[0000] Ensuring kernel weaveworks/ignite-kernel:4.19.47 exists, or importing it... 
DEBU[0000] Found kernel with UID 5dfb8abbb5d4c79a       
INFO[0000] Created VM with ID "030cd0a728543694" and name "test441" 
DEBU[0001] containerd: Inspecting image "weaveworks/ignite:v0.6.0-67-585d77f9804381" 
DEBU[0001] Writing "/var/lib/firecracker/vm/030cd0a728543694/runtime.containerd.resolv.conf" with new hash: "cca33446e2ce0f418633a6ea9ef9c54859cb8a755f90dcc74274200c31fa85e4", old hash: "" 
INFO[0001] Started Firecracker VM "030cd0a728543694" in a container with ID "ignite-030cd0a728543694" 

sudo bin/ignite exec test441 cat /etc/resolv.conf
#PROTO: DHCP
nameserver 1.0.0.1
bootserver 0.0.0.0

export vm_id="$(sudo bin/ignite ps | grep test441 | awk '{print $1}')"

ls -1 /var/lib/firecracker/vm/${vm_id}/*resolv.conf*
/var/lib/firecracker/vm/030cd0a728543694/runtime.containerd.resolv.conf
/var/lib/firecracker/vm/030cd0a728543694/runtime.containerd.resolv.conf.sha256

cat /var/lib/firecracker/vm/${vm_id}/*resolv.conf*
# The following config was built by ignite:
nameserver 1.0.0.1
search stealthybox.local
cca33446e2ce0f418633a6ea9ef9c54859cb8a755f90dcc74274200c31fa85e4

sudo ctr -n firecracker t exec --exec-id 0 "ignite-${vm_id}" cat /etc/resolv.conf
# The following config was built by ignite:
nameserver 1.0.0.1
search stealthybox.local

sudo bin/ignite vm stop test441 --log-level debug
INFO[0000] Removing the container with ID "ignite-030cd0a728543694" from the "cni" network 
INFO[0008] Stopped VM with name "test441" and ID "030cd0a728543694" 

sudo bin/ignite vm start test441 --log-level debug
DEBU[0000] containerd: Inspecting image "weaveworks/ignite:v0.6.0-67-585d77f9804381" 
DEBU[0000] "/var/lib/firecracker/vm/030cd0a728543694/runtime.containerd.resolv.conf" with hash "cca33446e2ce0f418633a6ea9ef9c54859cb8a755f90dcc74274200c31fa85e4" is unchanged 
INFO[0000] Networking is handled by "cni"               
INFO[0000] Started Firecracker VM "030cd0a728543694" in a container with ID "ignite-030cd0a728543694" 

sudo bin/ignite exec test441 cat /etc/resolv.conf
#PROTO: DHCP
nameserver 1.0.0.1
bootserver 0.0.0.0
```

### DIFF
--- a/pkg/resolvconf/resolvconf.go
+++ b/pkg/resolvconf/resolvconf.go
@@ -1,0 +1,109 @@
+package resolvconf
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"strings"
+
+	"github.com/weaveworks/ignite/pkg/util"
+
+	"github.com/miekg/dns"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	resolvDefault = "/etc/resolv.conf"
+	resolvSystemd = "/run/systemd/resolve/resolv.conf"
+)
+
+var (
+	// fallbackNameServers uses Google DNS
+	fallbackNameServers = []string{
+		"8.8.8.8",
+		"8.8.4.4",
+	}
+)
+
+// EnsureResolvConf ensures a valid, usable resolvConf is written to the filepath
+func EnsureResolvConf(filepath string, perm os.FileMode) error {
+	cfg, err := readDNSConfig()
+	if err != nil {
+		log.Warn(err)
+	}
+
+	filterLoopback(cfg)
+	// Fallback to default DNS servers
+	if len(cfg.Servers) == 0 {
+		for _, s := range fallbackNameServers {
+			cfg.Servers = append(cfg.Servers, s)
+		}
+	}
+
+	data := buildResolvConf(cfg)
+	err = util.WriteFileIfChanged(filepath, data, perm)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// readDNSConfig reads settings from /etc/resolv.conf -- if those settings indicate
+// systemd-resolved is in use, it reads them again from /run/systemd/resolve/resolv.conf.
+// If an error occurs, cfg will be defaulted to an empty dns.ClientConfig{}.
+func readDNSConfig() (*dns.ClientConfig, error) {
+	cfg, err := dns.ClientConfigFromFile(resolvDefault)
+	if err != nil {
+		return &dns.ClientConfig{}, fmt.Errorf("Using default DNS config: %v", err)
+	}
+
+	if isSystemdResolved(cfg) {
+		systemdCfg, err := dns.ClientConfigFromFile(resolvSystemd)
+		if err == nil {
+			return systemdCfg, nil
+		}
+	}
+
+	return cfg, nil
+}
+
+// isSystemdResolved returns whether the cfg likely indicate usage of systemd-resolved.
+// This function should be used with the ClientConfig from /etc/resolv.conf.
+func isSystemdResolved(cfg *dns.ClientConfig) bool {
+	if len(cfg.Servers) == 1 {
+		ip := net.ParseIP(cfg.Servers[0])
+		if ip != nil && ip.IsLoopback() && ip[len(ip)-1] == 53 {
+			return true // cfg.Servers is equivalent to ["127.0.0.53"] or ["::FFff:127.0.0.53"]
+		}
+	}
+	return false
+}
+
+// filterLoopback removes loopback addresses from cfg.Servers since they're
+// not usable as an upstream resolver for a vm
+func filterLoopback(cfg *dns.ClientConfig) {
+	servers := make([]string, 0)
+	for _, s := range cfg.Servers {
+		ip := net.ParseIP(s)
+		if ip != nil && !ip.IsLoopback() {
+			servers = append(servers, s)
+		}
+	}
+	cfg.Servers = servers
+}
+
+// buildResolvConf returns the bytes for a resolv.conf representing clientConfig
+func buildResolvConf(cfg *dns.ClientConfig) []byte {
+	s := "# The following config was built by ignite:\n"
+	if len(cfg.Servers) > 0 {
+		s += "nameserver " +
+			strings.Join(cfg.Servers, "\nnameserver ") +
+			"\n"
+	}
+	if len(cfg.Search) > 0 {
+		s += "search " + strings.Join(cfg.Search, " ") +
+			"\n"
+	}
+	return []byte(s)
+}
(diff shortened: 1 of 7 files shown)


pkg/runtime/containerd/client.go
```diff
@@ -376,7 +381,26 @@ func (cc *ctdClient) RunContainer(image meta.OCIImageRef, config *runtime.Contai
 	snapshotter := cc.client.SnapshotService(containerd.DefaultSnapshotter)
 
 	// Add the /etc/resolv.conf mount, this isn't done automatically by containerd
-	config.Binds = append(config.Binds, runtime.BindBoth("/etc/resolv.conf"))
+	// Ensure a resolv.conf exists in the vmDir. Calculate path using the vm id
+	// TODO(stealthybox):
```
stealthybox: This is here for posterity, but I can move it into an issue or just remove it altogether depending on what we want.

pkg/runtime/containerd/client.go
```diff
@@ -376,7 +381,26 @@ func (cc *ctdClient) RunContainer(image meta.OCIImageRef, config *runtime.Contai
...
+	//  - create snapshot ahead of time?
+	//    - is there a performance penalty for creating snapshot ahead of time?
+	//    - maybe we can use containerd.NewContainerOpts{} to do it just-in-time
+	//  - write this file to snapshot mount instead of vmDir
+	//  - commit snapshot?
+	//  - deprecate vm id from this function's signature
```
stealthybox: This extensive `TODO` is here for posterity, but I can move it into an issue or just remove it altogether depending on what we want.

pkg/resolvconf/resolvconf_test.go
```diff
@@ -0,0 +1,201 @@
+package resolvconf
```
stealthybox: This is its own package because it's very decoupled from runtime needs.
I couldn't find a more sensible home for it.

chanwit: Yes. I have no problem with this in a separate package :-)

pkg/util/fs.go
```diff
@@ -135,3 +138,38 @@ func FileIsEmpty(file string) (bool, error) {
 
 	return true, nil
 }
+
+// WriteFileIfChanged stores a sha of data at <filename>.sha256 and determines whether to
+// rewrite the file; it has the same signature as ioutil.WriteFile().
+func WriteFileIfChanged(filename string, data []byte, perm os.FileMode) error {
```
chanwit: I like this function!

pkg/resolvconf/resolvconf.go
```diff
@@ -0,0 +1,109 @@
...
+// systemd-resolved is in use, it reads them again from /run/systemd/resolve/resolv.conf.
+// If an error occurs, cfg will be defaulted to an empty dns.ClientConfig{}.
+func readDNSConfig() (*dns.ClientConfig, error) {
+	cfg, err := dns.ClientConfigFromFile(resolvDefault)
+	if err != nil {
+		return &dns.ClientConfig{}, fmt.Errorf("Using default DNS config: %v", err)
+	}
+
+	if isSystemdResolved(cfg) {
+		systemdCfg, err := dns.ClientConfigFromFile(resolvSystemd)
```
chanwit: This systemd bit could also solve my problem too.

Issue #229, opened 2019-07-19
Add Support to Display Information of a Running VM


robertojrojas:
It would be nice if one could run a command like:

`ignite ps --format "table {{.ID}}\t{{.Mounts}}"`

Similar to Docker's ps:

`docker ps --format "table {{.ID}}\t{{.Mounts}}"`

stealthybox:
+1
This would a great feature to support for all of our informational commands.

In the meantime here's a jq solution that might be helpful for somebody:
```shell
ignite ps -q | xargs -n1 ignite inspect vm \
  | jq -r '"\(.metadata.name) \n\t \(.status.ipAddresses) \t \(.spec.kernel.cmdLine)"'
```
```shell
rough-water 
	 ["172.18.1.70"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
dry-snowflake 
	 ["172.18.1.68"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
empty-resonance 
	 ["172.18.1.67"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
bitter-cherry 
	 ["172.18.1.69"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
ctrd_1 
	 ["172.18.1.63"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
hidden-dream 
	 ["172.18.1.66"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
small-wind 
	 ["172.18.1.65"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
crimson-river 
	 ["172.18.1.71"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
docker-cni 
	 ["172.18.1.64"] 	 console=ttyS0 reboot=k panic=1 pci=off ip=dhcp
```

stealthybox:
Also applies to `ignite vm ls`

Nit123:
Hello, I am a student from the University of Texas at Austin. I wanted to know if a group of my colleagues and I could work on this issue as part of an assignment for my virtualization class. Thanks in advance!

PR #458, opened 2019-09-24
Added: vm filtering

### LINKED ISSUE #228: Add Support to Filter VMs
It would nice if one could execute a command like:
    `ignite ps --filter <search>`

   example:
     `ignite ps --filter "label=k8s-cluster"`

This could work similarly to [Docker ps](https://docs.docker.com/engine/reference/commandline/ps/)

najeal:
This adds vm filtering to the `ignite ps` command with the `-f` or `--filter` flag.
We can now apply filtering on VM metadata
- Name
- UID
- Labels
- Annotations

We can use just one filter or multiple.
ex one filter
`sudo ignite ps --filter "{{.Name}}=my-vm"`
ex multi filters (separating them with a `,`):
- `sudo ignite ps --filter "{{.Name}}=my-vm",{{.UID}}=123`
- `sudo ignite ps --filter "{{.Name}}=my-vm",{{.Labels.mylabel}}=123`

Fixes #228

### DIFF
--- a/cmd/ignite/cmd/vmcmd/ps.go
+++ b/cmd/ignite/cmd/vmcmd/ps.go
@@ -21,6 +21,28 @@ func NewCmdPs(out io.Writer) *cobra.Command {
 		Long: dedent.Dedent(`
 			List all running VMs. By specifying the all flag (-a, --all),
 			also list VMs that are not currently running.
+			Using the -f (--filter) flag, you can give conditions VMs should fullfilled to be displayed.
+			You can filter on all the underlying fields of the VM struct, see the documentation:
+			https://ignite.readthedocs.io/en/stable/api/ignite_v1alpha2.html#VM.
+
+			Different operators can be used:
+			- "=" and "==" for the equal
+			- "!=" for the is not equal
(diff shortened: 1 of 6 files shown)


stealthybox:
Further U/X could be done in a follow-up:
- CLI `--help` Examples
- CLI `--help` list of valid fields for this command, or link to API docs -- mention user can see full object with `ignite inspect`
- matching on non Metadata fields:
`IMAGE, KERNEL, SIZE, CPUS, MEMORY, CREATED, STATUS, IPS, PORTS`

This is also related to the U/X of #229

najeal:
I think I'm ready for the next round

stealthybox:
Thanks so much for the persistence to get this through.
I'm playing with this locally -- the docs are very helpful.

This is a very high quality addition to ignite.

One tweak we can follow up with is to perform an `OR` operation when `-f` is passed multiple times. (this is what docker does)
We can do this by changing the flag type and running the filter multiple times.

We can also document `-f '{{.ObjectMeta.UID}}=092b26ee1196a114'`

This is great to merge as is :+1:

PR #460, opened 2019-09-26
Change default CNI network name, bridge name, and subnet

### LINKED ISSUE #424: Change the Ignite default CNI subnet to avoid collision with Docker 172.18.x.x
We now have an evidence to show that when setting the CNI bridge to be 172.18, it will eventually collide with a Docker bridge.

From @stealthybox investigation: when the collision happens, users see an message like:
```
INFO[0000] Created VM with ID "4420a64071626e72" and name "ctrd-421-stealthybox_1" 
INFO[0001] Started Firecracker VM "4420a64071626e72" in a container with ID "ignite-4420a64071626e72" 
FATA[0021] Tried connecting to SSH but timed out dial tcp 172.18.0.237:22: i/o timeout 
CNI-af6d5b7188e4eb5cd438df7e  all  --  172.18.0.237         anywhere             /* name: "ignite-containerd-bridge" id: "ignite-4420a64071626e72" */
ACCEPT     all  --  anywhere             172.18.0.0/16        /* name: "ignite-containerd-bridge" id: "ignite-4420a64071626e72" */
MASQUERADE  all  --  anywhere            !base-address.mcast.net/4  /* name: "ignite-containerd-bridge" id: "ignite-4420a64071626e72" */
FATA[0002] failed to dial: dial tcp 172.18.0.237:22: connect: no route to host 
```
Which caused by different route entry of the same subnet.
```
route | grep 172
172.17.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
172.18.0.0      0.0.0.0         255.255.0.0     U     0      0        0 br-f077d119eaa4
172.18.0.0      0.0.0.0         255.255.0.0     U     0      0        0 cni0
172.19.0.0      0.0.0.0         255.255.0.0     U     0      0        0 br-ca6c16b5ef35
```

We'll choose a more appropriate subnet for 0.7.0.

### LINKED ISSUE #455: Change the default cni0 bridge name to ignite0
Configuring this bridge will inconveniently collide with many CNI examples.
It's best to use our own bridge name.

Related #424

stealthybox:
Fixes #424
Fixes #455

The solution in this patch for #424 is to use a static subnet in the 10.0.0.0/8 range.
This avoids the complexity of collaborating with the docker defaulting logic which requires calculating and iterating a over a similar list of subnets and checking to see if they're in use by the kernel.
This:
- keeps the code simple
- has the future side-effect of making issues more searchable via google when users submit issue details with the default network setup

Rationale for the subnet choice is documented on the constant.

Changing these CNI details will ultimately will necessitate that any current users of our (mostly broken) default CNI setup will likely want to migrate.
This is not forced on users though. The config is still only written when the CNI net conf dir is empty.

### DIFF
--- a/pkg/network/cni/cni.go
+++ b/pkg/network/cni/cni.go
@@ -31,27 +31,42 @@ const (
 	CNIConfDir = "/etc/cni/net.d"
 	// netNSPathFmt gives the path to the a process network namespace, given the pid
 	netNSPathFmt = "/proc/%d/ns/net"
-	// igniteCNIConfName is the filename of Ignite's CNI configuration file
-	igniteCNIConfName = "10-ignite.conflist"
-	// igniteBridgeName specifies the default "docker-bridge"-like plugin for containerd to use if no other CNI plugin is available
-	igniteBridgeName = "ignite-containerd-bridge"
+
+	// defaultCNIConfFilename is the vanity filename of Ignite's default CNI configuration file
+	defaultCNIConfFilename = "10-ignite.conflist"
+	// defaultNetworkName names the "docker-bridge"-like CNI plugin-chain installed when no other CNI configuration is present.
+	// This value appears in iptables comments created by CNI.
+	defaultNetworkName = "ignite-cni-bridge"
+	// defaultBridgeName is the default bridge device name used in the defaultCNIConf
+	defaultBridgeName = "ignite0"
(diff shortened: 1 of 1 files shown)


stealthybox:
cc @najeal -- pinging you as discussed
Please review at your leisure :)

chanwit:
Thank you @najeal to help review this, and thank you @stealthybox for taking care of this.

lucidprogrammer:
Thanks for this update, however, its possibly causing #462

PR #461, opened 2019-09-26
Teardown IPMasq rules for all actual configured bridges instead of using the hardcoded default string

stealthybox:
Followup #460 

This logic seems to work well.
I added a trace line.

This protects us from a class of network removal bugs when stopping vm's that were created on the previous network names or CNI networks that are user configured.

You can create a vm with an old version of ignite on the `172.18/16` network, stop it with this new version without error, and then start it again. It will be on the new `10.61/16` network.

This gives us our migration plan.

### DIFF
--- a/pkg/network/cni/cni.go
+++ b/pkg/network/cni/cni.go
@@ -83,9 +83,10 @@ var defaultCNIConf = fmt.Sprintf(`{
 `, defaultNetworkName, defaultBridgeName, defaultSubnet)
 
 type cniNetworkPlugin struct {
-	cni     gocni.CNI
-	runtime runtime.Interface
-	once    *sync.Once
+	cni       gocni.CNI
+	cniConfig *gocni.ConfigResult
+	runtime   runtime.Interface
+	once      *sync.Once
 }
 
 func GetCNINetworkPlugin(runtime runtime.Interface) (network.Plugin, error) {
@@ -168,6 +169,9 @@ func (plugin *cniNetworkPlugin) initialize() (err error) {
 			log.Errorf("failed to load cni configuration: %v", err)
 		}
 	})
+
+	plugin.cniConfig = plugin.cni.GetConfig()
+
 	return
 }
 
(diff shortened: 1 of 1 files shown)


stealthybox:
+ @najeal

pkg/network/cni/cni.go
```diff
@@ -203,17 +207,30 @@ func (plugin *cniNetworkPlugin) RemoveContainerNetwork(containerID string) error
...
+				hasBridge = true
+			}
+		}
+
+		if hasBridge {
+			log.Debugf("TeardownIPMasq for container %q on CNI network %q which contains a bridge", containerID, net.Config.Name)
+			comment := utils.FormatComment(net.Config.Name, c.ID)
+			for _, t := range result {
+				if err = ip.TeardownIPMasq(t.ip, t.chain, comment); err != nil {
+					return err
```
chanwit: is it ok to do the immediate return here when fail?

or continue and collect the error in to an array, then check if the error size > 0?

I'm not sure which way is better because we're cleaning things here.

stealthybox: I'll wrap the list of errs and then Errorf

stealthybox:
@chanwit I consolidated the errors.

I moved this cleanup into function.
It's now called before fetching the netNSPath which means cleanup of the chains will still happen if the container no longer exists.

The error is wrapped with the util defer helper.

pkg/network/cni/cni.go
```diff
@@ -185,11 +189,18 @@ func cniToIgniteResult(r *gocni.CNIResult) *network.Result {
...
-func (plugin *cniNetworkPlugin) RemoveContainerNetwork(containerID string) error {
-	if err := plugin.initialize(); err != nil {
+func (plugin *cniNetworkPlugin) RemoveContainerNetwork(containerID string) (err error) {
+	if err = plugin.initialize(); err != nil {
 		return err
 	}
 
+	cleanupErr := plugin.cleanupBridges(containerID)
+	if cleanupErr != nil {
+		defer util.DeferErr(&err, func() error {
```
najeal: why do you defer the return error?

stealthybox: We can still try to remove the network if the cleanup fails.

stealthybox:
PR description edit:
> This protects us from a class of network removal bugs when stopping vm's that were created on the previous network names or CNI networks that are user configured.
> 
> You can create a vm with an old version of ignite on the 172.18/16 network, stop it with this new version without error, and then start it again. It will be on the new 10.61/16 network.
> 
> This gives us our migration plan.

PR #469, opened 2019-10-06
use ssh discover technique to check if SSH service is really run…

chanwit:
This PR improves the SSH wait mechanism by trying to authen the SSH server of a VM.
If the SSH returns "unable to authen" then it seems to be really up and running.

Thank you @lucidprogrammer who found this bug on his baremetal box.
I also found this bug too when running Ignite on a very machine.

@stealthybox this seems to be an urgent fix needed by tmr event.
and we also need it for v0.5.5.

### DIFF
--- a/cmd/ignite/run/start.go
+++ b/cmd/ignite/run/start.go
@@ -3,13 +3,14 @@ package run
 import (
 	"fmt"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/weaveworks/ignite/pkg/apis/ignite"
-
 	"github.com/weaveworks/ignite/pkg/operations"
 	"github.com/weaveworks/ignite/pkg/preflight/checkers"
 	"github.com/weaveworks/ignite/pkg/util"
+	"golang.org/x/crypto/ssh"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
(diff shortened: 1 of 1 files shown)


stealthybox:
This is unfortunately even slower than before.
It definitely is reasonable to see how this could address some SSH/`ignite exec` race conditions:

```bash
time sudo ignite-v0.6.1 run weaveworks/ignite-ubuntu --ssh
INFO[0000] Created VM with ID "352579ed0c948d0e" and name "broken-dew" 
INFO[0001] Started Firecracker VM "352579ed0c948d0e" in a container with ID "ignite-352579ed0c948d0e" 
sudo ignite-v0.6.1 run weaveworks/ignite-ubuntu --ssh  0.19s user 0.14s system 9% cpu 3.496 total

time sudo ./bin/ignite run weaveworks/ignite-ubuntu --ssh
INFO[0000] Created VM with ID "67548738f4251506" and name "blue-snowflake" 
INFO[0001] Started Firecracker VM "67548738f4251506" in a container with ID "ignite-67548738f4251506" 
sudo ./bin/ignite run weaveworks/ignite-ubuntu --ssh  0.21s user 0.15s system 6% cpu 5.751 total
```

PR #470, opened 2019-10-07
try split resize2fs min size with both colon and full with colon

chanwit:
This PR fixes a bug when using non-English locales.
Some locales use full width colon as a separator instead of the normal colon (:)

```
ignite create centos:7 --name lab2 --cpus 2 --memory 2GB --ssh
INFO[0000] Docker image "centos:7" not found locally, pulling...
INFO[0038] Starting image import...
FATA[0040] strconv.ParseInt: parsing "预计文件系统的最小尺寸：61817": invalid syntax
```


stealthybox:
Thanks!
Closing in favor of #473

PR #478, opened 2019-10-11
add new centos-firekube image skeleton

chanwit:


chanwit:
kindly ping @stealthybox 
this PR's ready to review.

stealthybox:
Notes from the [oct14 ignite dev call (recorded)](https://docs.google.com/document/d/1fv8_WD6qXfvlIq7Bb5raCGyBvc42dNF-l8uaoZzoUYI/edit#heading=h.3u1i1i4ncghi):
> - The technical approach for building this primed image appears technically sound
> - The yum cache is created in this image to workaround bugs that prevent the VM from booting when packages like docker-ce are pre-installed. This bug is not understood but may relate to SELinux.
> - This image build depends on a modified version of the unsupported ctr binary from containerd v1.2.9 that is capable of priming the containerd rootfs with images without unpacking them to a snapshotter implementation.
> - Maintaining this ctr129 binary will need to be done until we can use a docker/kubernetes version in Firekube that supports the brand-new containerd v1.3.0
> - This image is probably better managed in the Firekube project since it will need to be updated and released in when Firekube changes. Changes to ignite are not likely to cause changes to this image.
> - We should try hard not to have a separate image between the ignite and footloose implementations as that complicates the Firekube project.

PR #481, opened 2019-10-16
Ignite wrapper

chanwit:
This PR introduce `ignitew`, a self-contained binary bundling `ignite` and CNI plugin binaries together.

`ignitew` can be used as a drop-in replacement of `ignite` OOTB without user manually installing CNI plugins.

It uses the excellent `warp-packer` https://github.com/dgiagio/warp to create the self contained binary. It uses no sandbox, so that the behaviour could be expected to be similar to the normal `ignite`.

**Example:**
```
$ sudo ls -al /opt/cni/bin/bridge
ls: cannot access '/opt/cni/bin/bridge': No such file or directory
$ sudo ls -al /opt/cni/bin/loopback
ls: cannot access '/opt/cni/bin/loopback': No such file or directory
$ ./ignitew version
Ignite version: version.Info{Major:"0", Minor:"6+", GitVersion:"v0.6.0-108+472f5f5d9ca49f", GitCommit:"472f5f5d9ca49f63d55bf2433d
e7eeafc946787b", GitTreeState:"clean", BuildDate:"2019-10-16T12:19:41Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/am
d64"}
Firecracker version: v0.18.0
Runtime: containerd
$ sudo ./ignitew run weaveworks/ignite-ubuntu --ssh
INFO[0001] Created VM with ID "eee36d65bf8d132c" and name "little-forest" 
INFO[0001] Started Firecracker VM "eee36d65bf8d132c" in a container with ID "ignite-eee36d65bf8d132c" 
$ sudo ./ignitew exec little-forest -- uname -a
Linux localhost.localdomain 4.19.47 #1 SMP Tue Jul 16 18:57:23 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
$ sudo ./ignitew rm -f little-forest
INFO[0000] Removing the container with ID "ignite-eee36d65bf8d132c" from the "cni" network 
INFO[0000] Removed VM with name "little-forest" and ID "eee36d65bf8d132c" 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -265,6 +265,12 @@ serve-docs: build-docs
 	@echo Stating docs website on http://localhost:${DOCS_PORT}/_build/html/index.html
 	@$(DOCKER) run -i --rm -p ${DOCS_PORT}:8000 -e USER_ID=$$UID ignite-docs
 
+build-ignitew: build-all
+	$(DOCKER) build --no-cache -t ignitew-builder -f Dockerfile.ignitew .
+	$(DOCKER) create --name=ignitew-builder ignitew-builder
+	$(DOCKER) cp ignitew-builder:/output/ignitew .
+	$(DOCKER) rm -f ignitew-builder
+
 e2e: build-all e2e-nobuild
 
 e2e-nobuild:
--- a/pkg/preflight/checkers/checks.go
+++ b/pkg/preflight/checkers/checks.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	api "github.com/weaveworks/ignite/pkg/apis/ignite"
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -265,6 +265,12 @@ serve-docs: build-docs
 	@echo Stating docs website on http://localhost:${DOCS_PORT}/_build/html/index.html
 	@$(DOCKER) run -i --rm -p ${DOCS_PORT}:8000 -e USER_ID=$$UID ignite-docs
 
+build-ignitew: build-all
```
stealthybox: we can change this to `ignite` instead of `build-all`

pkg/preflight/checkers/checks.go
```diff
@@ -99,9 +100,20 @@ func StartCmdChecks(vm *api.VM, ignoredPreflightErrors sets.String) error {
 	for _, dependency := range constants.PathDependencies {
 		checks = append(checks, ExistingFileChecker{filePath: dependency})
 	}
+	// This should be warning only, in case users prefer other network plugins
 	if providers.NetworkPluginName == network.PluginCNI {
-		for _, dependency := range constants.CNIDependencies {
-			checks = append(checks, ExistingFileChecker{filePath: dependency})
+		// In case of using ignitew, we put CNI plugin binaries relatively to the executable
+		exeDir := filepath.Dir(os.Args[0])
+		if _, err := os.Stat(exeDir); err == nil {
```
stealthybox: I believe this may always be true -- it doesn't indicate that the binary is actually `ignitew`

Perhaps there's a warp native way to detect that the binaries have been unpacked for us

stealthybox:
The Makefile needs to be updated to output `ignitew` to the proper bin/arch directories with a symlink.

We should make note whether this also supports ARM64 builds.

stealthybox:
A build shows that the resulting binary size of `ignitew`:`ignite` is `51M` vs. `35M`.

This is likely just the size of the CNI binaries themselves. It's unfortunate because I believe every one of those is an independent Go binary :sweat_smile:

fire-ant:
This is a pretty cool all-in-one solution, has it just dropped to the bottom of the backlog?

Issue #449, opened 2019-09-20
pkg/runtime Bind{} should support read-only


stealthybox:
If you are in the `ignite-spawn` container, you have root and can change host files like `/etc/resolv.conf` using ignite `<= v0.6.1 || v.0.5.3`.

The `Bind` interface should be updated to support `ro`/`rw` or arbitrary mount options.
IMO specifying this should be required.

Nicholas227:
Hello, I am a student from the University Of Texas At Austin and as a part of my virtualization class assignment I wanted to ask if I would be allowed to work on this issue, I was wondering if I could be assigned to this issue. Thanks.

Nicholas227:
Hello, I am a student from the University Of Texas At Austin and as a part of my virtualization class assignment I wanted to ask if I would be allowed to work on this issue, I was wondering if I could be assigned to this issue. Thanks.

Nicholas227:
Hello, I am a student from the University Of Texas At Austin and as a part of my virtualization class assignment I wanted to ask if I would be allowed to work on this issue, I was wondering if I could be assigned to this issue. Thanks.

jerinthomas1404:
@stealthybox Can I take up this issue?

PR #483, opened 2019-10-25
Support arbitrary mount options

teaguecole:
PR for #449 to support arbitrary mount options. This is my first PR for Weaveworks Trying to start somewhere so I can get feedback and move in the right direction.


chanwit:
Thank you @teaguecole !

teaguecole:
Okay so this finally passes tests. I am not sure if I should be adding all the vendor files that get modified after running `make autogen` & `make tidy-in-docker`.  My gut instinct is that I was not suppose to and should be easy to undo.  @stealthybox or @chanwit Do you mind commenting on what I have done, and what I should do next?

pkg/runtime/containerd/client.go
```diff
@@ -504,10 +505,7 @@ func withMounts(binds []*runtime.Bind) oci.SpecOpts {
 			Source:      bind.HostPath,
 			Destination: bind.ContainerPath,
 			Type:        "bind",
-			Options: []string{
-				"rbind",
-				"rw",
-			},
+			Options: 	append(bind.MountOptions, "rw", "rbind"),
```
stealthybox: I suggest:
```
Options:	bind.MountOptions
```
No need to append rw,rwbind here.
This is just a conversion helper for going from ignite's `runtime.Bind{}` to containerd's `specs.Mount{}`.
The options should already be populated in the binds before calling this function.

stealthybox:
Nice work on this.
You've:
- updated our internal Bind interface to support options such as ro/rw
- updated our containerd runtime to support those mount options

You'll also want to add support for the Docker runtime.
See here:
https://github.com/weaveworks/ignite/blob/v0.6.2/pkg/runtime/docker/client.go#L133-L137

pkg/runtime/docker/client.go
```diff
@@ -133,7 +133,7 @@ func (dc *dockerClient) AttachContainer(container string) (err error) {
 func (dc *dockerClient) RunContainer(image meta.OCIImageRef, config *runtime.ContainerConfig, name, id string) (string, error) {
 	binds := make([]string, 0, len(config.Binds))
 	for _, bind := range config.Binds {
-		binds = append(binds, fmt.Sprintf("%s:%s", bind.HostPath, bind.ContainerPath))
+		binds = append(binds, fmt.Sprintf("%s:%s", bind.HostPath, bind.ContainerPath, bind.MountOptions))
```
stealthybox: Your editor is likely complaining about this line.
The format string does not match the number of arguments.

`bind.MountOptions` is a `[]string`, so you'd want to join them somehow.
This is one of the reasons it might be easier/cleaner to move to `Mounts` instead of `Binds` for docker's container spec.

teaguecole:
I got busy with other stuff, but I plan on working on this now.. should finish this up soon.

teaguecole:
This is ready for a review whenever anyone gets a chance. Ill probably hop on to the next dev meeting given that I am not traveling etc.

mdundek:
Hello guys,

What is happening with this `mount options` feature? Very interested in this, any idea when this will be available?

BR

teaguecole:
I have been relatively busy with work. I believe most of the work is done, just needs to be dusted off and rebased.  I am not sure when I will have time go about doing this.  Apologies for not keeping up with this.

PR #491, opened 2019-11-11
dependencies changed

chanwit:
This PR fixes the broken CI build & test as there are some dependency update via `go mod`.


stealthybox:
I ran the e2e and it looks like there are no regressions.
Docker+CNI removal is still acting up, but that's not because of these deps.

PR #494, opened 2019-11-13
Initial set of metrics for gitops processing

palemtnrider:
I put together a few simple metrics for GitOps processing.  Thoughts?


stealthybox:
This looks good -- I don't see anyone needing to control the metrics registrations as a dependency with the current shape of this interface, so the global vars seem appropriate.

PR #495, opened 2019-12-03
Add new command: `ignite cp`

### LINKED ISSUE #419: Copy files from host to VM, while the VM is in running state
Not able to copy while VM is running.
There is `--copy-files` flag, but only works with ignite run command

gpauloski:
This adds the command `ignite cp <vm> <source> <dest>` to copy files into a running VM.

Based on the `ignite ssh` implementation, `ignite cp` uses `scp` and implements the `-i`, `-t`, and `-r` flags.

Fixes #419


stealthybox:
@gpauloski, on the [ignite dev call (feb3,2020)](https://docs.google.com/document/d/1fv8_WD6qXfvlIq7Bb5raCGyBvc42dNF-l8uaoZzoUYI/edit#heading=h.2wbeu5jw5hk4) we're commenting on how useful this patch is.

@darkowlzz mentioned that he's using it for the kind integration, and that he's willing to refactor the feature with the go stdlib SSH package instead of shelling out to the host's `ssh` binary in the current mountNamespace.

I mentioned we should make sure to preserve your git history as he extends your patch.
Does this sound good to you?

Thanks for building this.

gpauloski:
@stealthybox Absolutely, sounds good to me.

stealthybox:
@gpauloski Thanks so much for starting this work.
Your commits have been preserved in the history of #550.
We are planning to merge that patch.
The work you started here will be great for users :)

PR #497, opened 2019-12-05
Add loadable module for TCMU-based storage devices

croomes:
This change adds the `tcm_core_user` module to the kernel image.

`tcm_core_user` is the module providing [TCMU](https://www.kernel.org/doc/Documentation/target/tcmu-design.txt), used by storage systems like Ceph, Gluster, Longhorn. OpenEBS and StorageOS.  This will allow Pods running in Firekube to mount PVCs provisioned by these storage products.

We've been using ignite with a custom kernel for testing StorageOS but Firekube opens this up to a much wider audience, hence wanting to get this in the base image.


croomes:
The test fail looks unrelated but I'm not able to re-run the test to see if was a flake.

croomes:
@twelho @dholbach does this look ok?

dholbach:
@chanwit and @stealthybox can probably help.

stealthybox:
@croomes, I'm not a subject matter expert on this module, but I think this fits well into a kernel intended for kubernetes metal use.

Heads up that this may need some small edits if #506 / #511 merges first.

stealthybox:
:wave: Thanks for this contribution @croomes 

Would you be able to add and document this to the `images/kernel/config-patches`?
Running the build with that change on a fresh copy of `master` will update all of the kernel configs.

Let me know if you don't have the bandwidth.
@darkowlzz or I can carry it :)

darkowlzz:
Added the TCM_USER2 config in config-patches and ran patch.config.sh to update the generated configs. Looks like the generated config files were not up-to-date with the config-patches and versioned configs. This resulted in many updates to the generated configs in addition to the TCMU configs.

stealthybox:
Thanks @croomes & @darkowlzz

PR #503, opened 2019-12-12
Added stream contents of FIFO files to logs

### LINKED ISSUE #88: Save the contents of the FC fifos to files
Firecracker emits one fifo for metrics and one for VMM logs. Both of these are by nature very ephemeral. `ignite-spawn` should read the fifos periodically (or better, stream) and write the contents to persistent log files in `/var/lib/firecracker/vm/$id/{vmm,metrics}.log`

gpauloski:
This PR adds goroutines launched in `ExecuteFirecracker()` which watch the FIFO files emitted by Firecracker and streams them to a persistent file with the same name but ending in `.log`.

We add the function `NamedPipeWatcher(pipePath string, filePath string)` in ` pkg/util/util.go` which handles streaming the contents from the FIFO at `pipePath` to a file at `filePath`.

If there is a more appropriate place for this function we can move it.

Fixes #88

### DIFF
--- a/pkg/container/firecracker.go
+++ b/pkg/container/firecracker.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/signal"
 	"path"
+	"runtime"
 	"strconv"
 	"syscall"
 	"time"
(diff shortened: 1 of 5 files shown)


pkg/container/firecracker.go
```diff
@@ -136,6 +137,16 @@ func ExecuteFirecracker(vm *api.VM, dhcpIfaces []DHCPInterface) error {
 
 	installSignalHandlers(ctx, m)
 
+	// Create file names for persistent files to save FIFO stream to
+	logSocketPathPersistent := path.Join(vm.ObjectPath(), constants.LOG_FIFO_PERSISTENT)
+	metricsSocketPathPersistent := path.Join(vm.ObjectPath(), constants.METRICS_FIFO_PERSISTENT)
+
+	runtime.GOMAXPROCS(3)
```
stealthybox, line 144: The rebase on this patch is looking healthy and passing tests.

@gpauloski, do you still remember why you added this change to GOMAXPROCS and MAX_CONCURRENT_WRITERS ?

If we could justify these things in some comments, that would be very excellent.

Thanks for your patience on this PR review.
This is a very desirable feature :)

gpauloski: Apologies for the very overdue response. I do not recall why I added this change but it seems strange---maybe an artifact of being new to Go at the time.

I'm going to close this since its been a while :).

PR #506, opened 2020-01-20
Completely restructure & document the kernel image build

luxas:
Also adding the latest LTS

cc @chanwit @stealthybox


luxas:
ping @stealthybox @chanwit please merge this PR, so we can unblock #511

images/kernel/Makefile
```diff
@@ -1,11 +1,22 @@
-KERNEL_VERSION ?= 4.19.47
-KERNEL_EXTRA ?= 
+# Check https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/refs/ for updates
+KERNEL_VERSIONS ?= 4.14.166 4.19.97 5.4.13
```
stealthybox: Noting that these will not overwrite the old images @ https://hub.docker.com/r/weaveworks/ignite-kernel/tags since they are all version bumps :+1:

images/kernel/upgrade-config.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+KERNEL_BUILDER_IMAGE=luxas/kernel-builder:gcc-7-2
```
stealthybox: This is not a publicly accessible image tag: https://hub.docker.com/r/luxas/kernel-builder/tags
Did you mean `gcc-7-2019-02-06` or `gcc-7` ?

images/kernel/patch-config.sh
```diff
@@ -0,0 +1,61 @@
...
+
+for file in ./versioned/*; do
+    old_file=${file}
+    new_file="./generated/$(basename ${file})"
+    
+
+    # Copy the old config file to the new (overwrite if present), and patch the new one in-place
+    cp ${old_file} ${new_file}
+    # Add an extra newline to the upstream file if it hasn't got it
+    echo "" >> ${new_file}
```
stealthybox:
```suggestion
    tail -c1 file | read -r _ || echo >> "${new_file}"
```
ref: https://backreference.org/2010/05/23/sanitizing-files-with-no-trailing-newline/

images/kernel/upgrade-config.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [[ $# != 2 ]]; then
+    echo "Usage: $0 [FROM] [TO]"
+    exit 1
+fi
+
+FROM=$1
+TO=$2
+VERSION=$(echo ${TO} | rev | cut -d- -f1 | rev)
```
stealthybox: Nice trick here! -- suggesting a comment:
```suggestion
VERSION="$(echo ${TO} | rev | cut -d- -f1 | rev)"  # Extracts the trailing hyphenated field -- this is dependent on naming the resulting file properly (ex: ./versioned/config-amd64-5.4.13)
```

images/kernel/README.md
```diff
@@ -0,0 +1,41 @@
...
+- 4.14.x
+- 4.19.x
+- 5.4.x
+
+The exact patch versions may be found in the [Makefile](Makefile).
+The available versions exist in the [stable kernel git tree](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/refs/).
+
+## Upgrading to a new kernel version
+
+It should be as easy as:
```
stealthybox: It's best not to suggest something is easy/obvious in our docs.
This is a pretty magical command -- it's not simple to determine what it is doing and what the necessary order of operations is for somebody to maintain these configs and directories.

```suggestion
The kernel Makefile has an `upgrade` command that will generate patched kernel configs for each specific version in `KERNEL_VERSIONS`.

The linux kernel source code is checked out in a build container for each target version, and a resulting "defconfig" based on each Firecracker recommended base-config in the `upstream/` directory is then output to the `generated/` directory.

Once `make upgrade-%` has generated a firecracker base-config for all `KERNEL_VERSIONS`, the resulting generated configs are copied to the `versioned/`dir and are patched with ignite specific `./config-patches`. These configs under `versioned/` are used for the matching kernel builds.

Run:
```

images/kernel/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+```console
+$ make upgrade
+```
+
+after you've upgraded the values in the Makefile.
```
stealthybox:
```suggestion
after you've set new `KERNEL_VERSIONS` in the Makefile or changed `./config-patches`.
This will regenerate and produce new patched, versioned configs.

Older kernel versions should be deleted from `generated/` and `versioned/` when checking in new ones.
```

images/kernel/Makefile
```diff
@@ -1,11 +1,22 @@
...
-	docker build -t weaveworks/ignite-kernel:${KERNEL_VERSION}${KERNEL_EXTRA} \
-		--build-arg KERNEL_VERSION=${KERNEL_VERSION} \
-		--build-arg KERNEL_EXTRA=${KERNEL_EXTRA} .
 
-push:
-	docker push weaveworks/ignite-kernel:${KERNEL_VERSION}${KERNEL_EXTRA}
+upgrade: $(addprefix upgrade-,$(KERNEL_VERSIONS))
+	./patch-config.sh
+	for file in generated/*; do \
+		./upgrade-config.sh $$file $$file; done
```
stealthybox: Can you help me understand why this is here?
I'm reading that this runs
```bash
     make clean && make mrproper && cp /tmp/.config . && \
     make EXTRAVERSION="" LOCALVERSION= olddefconfig
```
on the already generated config as the input.
That output then overwrites the input file.

Does this cleanup or do something to the generated config that changes it?
Those changes wouldn't end up in the patch configs unless you ran `make upgrade` twice.

Thanks :+1:

images/kernel/patch-config.sh
```diff
@@ -0,0 +1,61 @@
...
+        echo "    Applying: ${config_key}=${config_value}"
+        if [[ ${config_value} == "n" ]]; then
+            unset_kernel_config ${config_key} ${config_file}
+        else
+            set_kernel_config ${config_key} ${config_value} ${config_file}
+        fi
+    done
+}
+
+for file in ./versioned/*; do
```
stealthybox: Perhaps this loop on all versioned configs should be moved to a loop in the `Makefile`.
That loop could be written to only include files that are in the intended `KERNEL_VERSIONS`.

stealthybox:
Thanks for building this -- it's really nicely done.

images/kernel/Makefile
taqtiqa-mark: Apologies in advance ... I'm just getting my head around Ignite - by spec'ing and writing a Cheff Cookbook to install `ignite and ignited`:
Is there a Linux kernel dependency to take into consideration when installing Ignite.

stealthybox: @taqtiqa-mark these kernel versions are for the VM guests that ignite runs. They're distributed as OCI images so you don't have to worry about installing them.

For installing ignite, you need a linux host capable of meeting Firecrackers minimum runtime dependencies.
Currently, that is a host with Linux Kernel 4.14 or newer and KVM enabled:
- https://github.com/firecracker-microvm/firecracker/blob/master/FAQ.md#what-operating-systems-are-supported-by-firecracker

Please see our documented requirements:
- https://ignite.readthedocs.io/en/stable/dependencies.html#requirements-and-dependencies

images/kernel/README.md
luxas: Good pointer, thanks!

images/kernel/Makefile
luxas: This basically sanitizes, reorganizes and formats (overwrites) the kernel config for the final generated configs

images/kernel/patch-config.sh
luxas: Future improvements IMO, let's get ARM support merged first

luxas:
Ready to merge!
The CI failure is unrelated, and can/should be fixed in an other PR

PR #512, opened 2020-01-28
Don't fail preflight when only `docker-containerd` is available

luxas:
In the `docker.io` package in Debian 10, the containerd binary is shipped as `docker-containerd`
At the moment, Ignite fails the preflight checks for this. This PR fixes this false negative.
Similarly, if `docker-containerd` is available (only), then we should fall back to `RuntimeLinuxV1` without issuing the extra warning (not needed, it works well).

This PR also propagates through the ignore-preflight-checks flag to the user, and fixes minor bugs related to this.

@stealthybox @chanwit Please review


stealthybox:
```shell
make ignite bin/amd64/Dockerfile e2e-nobuild \
  E2E_REGEX="'DockerAndDockerBridge|Containerd'"
```
e2e passes ^

Verified the flags are added to `ignite run/start` now
Thanks for fixing that :+1:

PR #515, opened 2020-01-30
Cleanup container networking on vm rm

### LINKED ISSUE #450: `ignite vm rm` doesnt cleanup CNI
Follow-up to #442 

@chanwit and I tested that the `CNI-FORWARD` rules per-vm are cleaned up on `ignite vm stop`.

However, I noticed this doesn't happen on `ignite vm rm`.
We could also check CNI state on vm removal and remove if necessary.

michaelbeaumont:
Fixes #450


chanwit:
Thank you @michaelbeaumont for this fix!

PR #516, opened 2020-02-02
Add labels to VMs

darkowlzz:
Adds a new flag `--label` to VM create command for setting label on the VM
object. The flag is of StringArray type and can be used multiple times
to set more than one label.

```
# ignite create weaveworks/ignite-ubuntu --name my-vm1 -l label1=val1 -l label2=val2
```
vm inspect result:
```
# ignite inspect vm my-vm1
{
  "kind": "VM",
  "apiVersion": "ignite.weave.works/v1alpha2",
  "metadata": {
    "name": "my-vm1",
    "uid": "70d265ddaaa92462",
    "created": "2020-02-02T12:37:21Z",
    "labels": {
      "label1": "val1",
      "label2": "val2"
    }
...
```

### DIFF
--- a/pkg/metadata/metadata.go
+++ b/pkg/metadata/metadata.go
@@ -1,10 +1,12 @@
 package metadata
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path"
 	"regexp"
+	"strings"
 
 	"github.com/weaveworks/gitops-toolkit/pkg/filter"
 	"github.com/weaveworks/gitops-toolkit/pkg/runtime"
@@ -21,6 +23,9 @@ var (
 	uidRegex  = regexp.MustCompile(`^[a-z0-9]{16}$`)
 )
 
+// ErrNilObject is returned when the runtime object being initialized is nil.
+var ErrNilObject = errors.New("object cannot be nil when initializing runtime data")
+
 type Metadata interface {
 	runtime.Object
 }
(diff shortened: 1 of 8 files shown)


pkg/metadata/metadata.go
```diff
@@ -44,6 +49,23 @@ func SetNameAndUID(obj runtime.Object, c *client.Client) error {
...
 }
 
+// SetLabels metadata labels for a given object.
+func SetLabels(obj runtime.Object, labels []string) error {
+	if obj == nil {
+		return ErrNilObject
+	}
+
+	for _, label := range labels {
+		kv := strings.Split(label, "=")
```
chanwit: should we use `SplitN(label, "=", 2)`, wdyt?
if so, the value could also be possible to contain other = chars too.

darkowlzz: Yes, that's a good idea. I checked what happens if I pass "k1=v1,k2=v2" as label to docker, and it splits at first separator and accepts the rest of the string as the value.
```
            "Labels": {
                "k1": "v1,k2=v2"
            }
```
Updating the test to not fail for this case.
Thanks!

darkowlzz:
Added another check to ensure that the label name isn't empty. Updated the test with this case.

chanwit:
Thank you @darkowlzz for this wonderful PR!!

PR #517, opened 2020-02-03
Fix false positive object already exists error

darkowlzz:
`verifyUIDOrName()` performs a prefix match of a given object name. This
results in false positive errors when objects have similar names.

For example, if an object `vm11` already exists. Creating another object
`vm1` results in already exists error.

Instead of `NewIDNameFilter()`, use `NewNameFilter()` for exact match
filter.

Adds tests to verify the fix.


stealthybox:
+1
/lgtm

PR #518, opened 2020-02-04
Template formatted output for ps & inspect

darkowlzz:
Adds new flag `--template` (`-t`) for Go template formatted output in `inspect` and `ps` subcommands.
Since `ps` command already has a `-f` flag for `--filter`, this uses `-t` for template formatted output.

Adds golden files based tests for Inspect() and Ps() to verify the results.

### DIFF
--- a/cmd/ignite/cmd/inspect.go
+++ b/cmd/ignite/cmd/inspect.go
@@ -42,4 +42,5 @@ func NewCmdInspect(out io.Writer) *cobra.Command {
 
 func addInspectFlags(fs *pflag.FlagSet, i *run.InspectFlags) {
 	fs.StringVarP(&i.OutputFormat, "output", "o", "json", "Output the object in the specified format")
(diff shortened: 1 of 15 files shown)


darkowlzz:
Updated the commits to replace separate flags for updating the test golden files with a single `-update` flag for all the test updates.

stealthybox:
The template format works great, but the golden-file tests are flaky because the output order is not sorted and stable.

Running with `-count 1` will disable test caching.
Try `-count 10` as well:
```
go test ./cmd/ignite/run -count 10 -run TestPs
```

Perhaps we should consider sorting by name or ID before outputting the table.
We don't need to fix this in this PR.

darkowlzz:
Thanks for catching the issue.
Rebased the changes on latest master.
Changed the list of existing VMs to use a slice of `runtime.ObjectMeta`, instead of a map of VM name and ID, and tested it with `-count=10`. Seems to work reliably now.

PR #525, opened 2020-02-17
rm: Add --config flag

darkowlzz:
Adds `--config` flag in rm command to accept VM removal using the same
config file that was used to create the VM. VM name or UID is required
when passing a config file.

Adds tests to verify the changes.

### DIFF
--- a/cmd/ignite/run/rm.go
+++ b/cmd/ignite/run/rm.go
@@ -4,12 +4,15 @@ import (
 	"fmt"
 
 	api "github.com/weaveworks/ignite/pkg/apis/ignite"
+	"github.com/weaveworks/ignite/pkg/apis/ignite/scheme"
 	"github.com/weaveworks/ignite/pkg/operations"
 	"github.com/weaveworks/ignite/pkg/providers"
 )
 
+// RmFlags contains the flags supported by the remove command.
 type RmFlags struct {
-	Force bool
+	Force      bool
+	ConfigFile string
 }
 
 type rmOptions struct {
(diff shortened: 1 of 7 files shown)


cmd/ignite/run/rm.go
```diff
@@ -4,25 +4,55 @@ import (
...
+		// Name or UID must be provided in the config file.
+		if len(vm.Name) == 0 && len(vm.UID) == 0 {
+			return ro, fmt.Errorf("API resource config must have Name or UID")
+		}
+		ro.vms = []*api.VM{vm}
+		return ro, nil
+	}
+
+	// Use vm args to find the VMs to be removed.
+	if len(vmMatches) < 1 {
```
chanwit: I see. You removed minimumNArgs(1) above because you added a check here.

cmd/ignite/run/rm.go
```diff
@@ -4,25 +4,55 @@ import (
...
 
-func (rf *RmFlags) NewRmOptions(vmMatches []string) (ro *rmOptions, err error) {
-	ro = &rmOptions{RmFlags: rf}
+// NewRmOptions creates and returns rmOptions with all the flags and VMs to be
+// removed.
+func (rf *RmFlags) NewRmOptions(vmMatches []string) (*rmOptions, error) {
+	ro := &rmOptions{RmFlags: rf}
+
+	// If config file is provided, use it to find the VM to be removed.
+	if len(rf.ConfigFile) != 0 {
```
chanwit: if it's provided, you mean `if len(rf.ConfigFile) > 0` ?

cmd/ignite/run/rm_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	api "github.com/weaveworks/ignite/pkg/apis/ignite"
+	"github.com/weaveworks/ignite/pkg/apis/ignite/scheme"
+	meta "github.com/weaveworks/ignite/pkg/apis/meta/v1alpha1"
+	"github.com/weaveworks/ignite/pkg/client"
+	"github.com/weaveworks/ignite/pkg/providers"
+	"github.com/weaveworks/ignite/pkg/util"
+)
+
+func TestNewRmOptions(t *testing.T) {
```
chanwit: This test LGTM

chanwit:
Thank you @darkowlzz for this PR.
There's a small nit and I'm also not sure why the test is failing. It might be unrelated.

cmd/ignite/run/rm.go
darkowlzz: That's right, I was checking if config file path isn't empty. In the lines of https://github.com/weaveworks/ignite/blob/v0.6.3/cmd/ignite/run/create.go#L73
Could have compared with `""` :slightly_smiling_face:  
Should I still change it?

cmd/ignite/run/rm.go
darkowlzz: Yes, that's right. When a config file is provided, a vm argument is not necessary.

cmd/ignite/run/rm_test.go
darkowlzz: Thanks :slightly_smiling_face:

cmd/ignite/run/rm.go
chanwit: np.

chanwit:
@darkowlzz @stealthybox
The dependency changed and it failed the CI. It needs `make tidy` a bit.
But I'll be going to merge this if you guys are OK with it?

darkowlzz:
In the last meeting, we discussed about the CI failures due to dependency update check. We talked about removing the dependency update check. I'm okay with merging this and removing the check separately.

chanwit:
LGTM then :-)

Issue #528, opened 2020-02-24
Documentation: Integer constraint - diskSize: [size]


bbros-dev:
[Current declarative documentation](https://ignite.readthedocs.io/en/stable/declarative-config.html) makes no mention of the integer constraint  
```yaml
  # Optional, how much free writable space the VM should have at runtime
  # Default: 4GB
  diskSize: [size]
```

However:

```bash
# ignite run --config /etc/firecracker/manifests/smoke-test.yml
FATA[0000] v1alpha2.VM.Spec: v1alpha2.VMSpec.Image: v1alpha2.VMImageSpec.DiskSize: unmarshalerDecoder: strconv.Unmarsh
alText: parsing "1.23 GB": invalid syntax, error found in #10 byte of ...|:"1.23 GB","image":{|..., bigger context ...
|1e60f3b2f5"},"spec":{"cpus":1,"diskSize":"1.23 GB","image":{"oci":"weaveworks/ignite-ubuntu"},"memor|..
```

bbros-dev:
Syntax error: no spaces allowed between 1.23GB

stealthybox:
I can reproduce this issue.
My gut is that we should fix the parser.

bbros-dev:
Yeah, I'm leaving this to your unit test suites, and rather keep the integration tests at the integration of:

1. daemon
1. service manager
1. firecracker

bbros-dev:
Closing.  Following a review, we will not be using ignite outside of developer personal workflows, so won't be able to devote any time to issues outside of that scope.

PR #538, opened 2020-02-27
Documentation: closes issue #528

bbros-dev:


bbros-dev:

Issue #529, opened 2020-02-24
Documentation: uid format


bbros-dev:
[Current declarative documentation](https://ignite.readthedocs.io/en/stable/declarative-config.html) makes no mention of UID required format `^[a-z0-9]{16}$`

bbros-dev:
Closing.  Following a review, we will not be using ignite outside of developer personal workflows, so won't be able to devote any time to issues outside of that scope.

PR #539, opened 2020-02-27
Documentation: Closes issue #529

bbros-dev:
Documentation: Closes issue #529


bbros-dev:

Issue #549, opened 2020-03-02
Device /dev/ptp0 not available by default (c.f. upstream)


bbros-dev:
[Here Ignite indicates](https://github.com/weaveworks/ignite/blob/edfdea71fe14e56912fb402b337522dd2db71cef/images/kernel/config-patches#L14) it has deviated from upstream (Firecracker) to avoid the timing issue discussed [here upstream](https://github.com/firecracker-microvm/firecracker/blob/master/FAQ.md#my-guest-wall-clock-is-drifting-how-can-i-fix-it).

However, it doesn't appear to have the desired effect or am I missing something:

```bash
# dmesg | head -1
[    0.000000] Linux version 4.19.47 (root@b077421cb135) (gcc version 7.4.0 (Ubuntu 7.4.0-1ubuntu1~18.04
.1)) #1 SMP Tue Jul 16 18:57:23 UTC 2019

# apt install linuxptp
...
Preparing to unpack .../linuxptp_1.8-1_amd64.deb ...
Unpacking linuxptp (1.8-1) .................................................] 
Setting up linuxptp (1.8-1) ...################.............................] 
Created symlink /etc/systemd/system/multi-user.target.wants/ptp4l.service → /lib/systemd/system/ptp4l.se
rvice.
/usr/sbin/policy-rc.d returned 101, not running 'start ptp4l.service'
root@localhost:~#
root@localhost:~# ptp4l -i eth0 -m -q
ptp4l[1987.535]: interface 'eth0' does not support requested timestamping mode
failed to create a clock
root@localhost:~# ptp4l -i eth0 -p /dev/ptp0
failed to create a clock
```

stealthybox:
@bbros-dev, are you experiencing this with a build of your own kernel from ignite master?

In order to do so, you'd need to run a VM with your own `--kernel-image` set.

We haven't published builds of the new kernel in master yet as we still have some things to work out for ARM support.

bbros-dev:
Apologies for not posting earlier.
From `ignite vm` -> `weaveworks/ignite-kernel:4.19.47`

bbros-dev:
Closing.  Following a review, we will not be using ignite outside of developer personal workflows, so won't be able to devote any time to issues outside of that scope.

bbros-dev:
Re-opening this because we think the cause due to the logic in one of the kernel build scripts - **the correct config variable values do seem to be in the patch config file**.

Specifically, an initial reading suggests the root cause is this line:

[...]

PR #551, opened 2020-03-03
Issue 549: Device /dev/ptp0 not available

bbros-dev:
This takes one step toward resolving issue 549.

The step is to reflect what Weaveworks considers the upstream kernel config the weaveworks kernel builds start from.
These changes to upstream could be done by checking in patch file(s) that can simply be applied to the upstream kernel config file - vaporware.

Does not alter Weaveworks builds.


stealthybox:
Related #549

bbros-dev:
@stealthbox, no objection to updating to that 'version' of the upstream kernel config?
I just grabbed the latest in the absence of any criteria.

bbros-dev:

PR #552, opened 2020-03-04
e2e test: Run the tests with vendored deps

darkowlzz:
Since the dependencies are checked into the repo, use vendored
dependencies while running e2e tests.


chanwit:
LGTM

PR #553, opened 2020-03-05
Tiny image import fix

### LINKED ISSUE #402: Cant pull Docker hello-world image
image

darkowlzz:
When an image is very small, filesystem creation using mkfs fails due to
not enough space. This adds a minimum base image size that's used when
the image is too small.

Also, fixes the /etc/resolv.conf symlink error when there's no /etc
directory in the image filesystem. /etc dir is created as part of the
import step if it doesn't exists.

Adds an e2e test for this fix using hello-world docker container image
which is tiny and lacks /etc dir.

Fixes #402

### DIFF
--- a/e2e/image_test.go
+++ b/e2e/image_test.go
@@ -1,0 +1,33 @@
+package e2e
+
+import (
+	"fmt"
+	"os/exec"
+	"testing"
+
+	"gotest.tools/assert"
+)
+
+func TestImportTinyImage(t *testing.T) {
+	assert.Assert(t, e2eHome != "", "IGNITE_E2E_HOME should be set")
+
+	// NOTE: Along with tiny image, this also tests the image import failure
+	// when there's no /etc directory in the image filesystem.
+
+	testImage := "hello-world:latest"
+	// Remove if the image already exists.
+	rmvImgCmd := exec.Command(
+		igniteBin,
+		"image", "rm", testImage,
+	)
+	// Ignore error if the image doesn't exists.
+	_, _ = rmvImgCmd.CombinedOutput()
+
+	// Import the image.
+	importImgCmd := exec.Command(
+		igniteBin,
+		"image", "import", testImage,
+	)
+	importImgOut, importImgErr := importImgCmd.CombinedOutput()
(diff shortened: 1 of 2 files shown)


bbros-dev:
From your comment in #402 

>I'm not sure about a solution for this. Maybe just log a warning when resolv.conf setup fails?

Can you document the choice.

darkowlzz:
> Can you document the choice made, either in a FAQ type section hinting at the cause of certain errors or in the kernel build documentation?

This will not be a problem anymore after this is merged. This is following the firecracker networking option [Kernel Boot Param](https://github.com/firecracker-microvm/firecracker-containerd/blob/master/docs/networking.md#a-preferred-kernel-boot-parameters) . Ignite image import was failing to create a symlink from `/etc/resolv.conf` to `/proc/net/pnp` because images like `hello-world` and `busybox` don't have `/etc` directory at root directory. With this fix, the import step will create `/etc` if it doesn't exists.

bbros-dev:
> Testing one of them should be enough

Apologies for not being clearer.  I didn't suggest testing both. Just the real world one.

darkowlzz:
@bbros-dev  As per my understanding, this issue is about two fixes, small size and missing /etc dir. Busybox is about ~1.2 MB, which is more than what we need to test the fix. `hello-world` image is ~1.8 kB, which fits the need for this small size test. Dockerfile source for both [busybox](https://github.com/docker-library/busybox/blob/9d1c74a415de8f8795dfe3cd42b97c9ae7ef770f/uclibc/Dockerfile) and [hello-world](https://github.com/docker-library/hello-world/blob/7ecae6978055d2fb6960e2a29d24a2af612e2716/amd64/hello-world/Dockerfile) are not very different.
If you can recommend any other real image that's as small as `hello-world` (requirement: image size * 3 < 500 kB) we can use that.

bbros-dev:
No worries.

stealthybox:
Regardless of the lack of a known example for a vm userspace < 0.5MB, this fix is very reasonable.
Thanks all for taking the time to report, discuss, and implement it.

Issue #315, opened 2019-08-12
Support ComponentConfig?


luxas:
As `ignited` grows in complexity, instead of registering tons of flags on it, we'd use a declarative configuration file, aka. ComponentConfig, for it. (Could be even backed by git?)

stealthybox:
Some defaults and values that could be configurable:
```yaml
- runtime
- network-plugin
- CNI subnet
- CNI options
- CNI network name
- docker network name
- dns.nameservers[]  (not implemented)
- dns.search[]  (not implemented)
- ipv6 support  (not implemented)
- vm.cpus
- vm.size
- vm.memory
- log-level
- kernel.image
- kernel.args
- gitops.repo-url
- gitops.branch / ref
- gitops.paths
```
These can be overridden by flags or the appropriate field in an Ignite VM API object.

stealthybox:
Just brainstorming:
```shell
ignited daemon --config production-config.yaml
```
```yaml
apiVersion: ignite.weave.works/v1alpha1
kind: Configuration
metadata:
  name: production-config
spec:
  logLevel: debug
  runtime: docker
  networkPlugin: docker-bridge
  dockerBridge:
    network: testnet
  cni:
    subnet: 10.39.0.0/14
    networkName: prod-ignite-vm-bridge
    # options: 
  dns:
    # ipv6: true
    nameservers:
      - 1.1.1.1
      - 9.9.9.9
    search:
      - deployment.example.com
      - production.example.com
  gitops:
    repoURL: git@github.com/stealthybox/example
    ref: master
    paths:
      - ignite/production-backend
      - ignite/production-cache
  kernel:
    image: <something>
    cmdLine:
      - my
      - default
      - kernel
      - args
    vm:
      cpus: 4
      size: 20GB
      memory: 4GB
```

stealthybox:
For the `ignite` tool since we do not yet have a complete daemon architecture, we could load this configuration file from some well known paths:

`/etc/ignite/runtime.yaml`
related XDG_CONFIG_HOME for root user?

bbros-dev:
I'd say this belongs in XDG_DATA_HOME: `~/.local/share/ignite`, and the XDG_CONFIG_HOME would contain the much simpler configuration defaults such as:
```yaml
manifest:
  paths: 
    - /etc/firecracker/manifests
    - /dfs/db/firecracker/manifests
    - /dfs/sec/firecracker/manifests
    - /home/me/.local/share/ignite/manifests
[...]

PR #555, opened 2020-03-09
[POC] Add ignite configuration file support

darkowlzz:
Defines APIs for ignite configurations with `Configuration` object.

Adds a new persistent flag at the root command, `ignite-config`.
This flag can be used to set the file path of an ignite config file.

An ignite config file contains the ignite Configuration object. It can
be used to set the default configurations of various ignite components.
By default, if a config file is found in /etc/ignite/config.yaml, it is
used to configure ignite. `--ignite-config` flag can be used to override
any default config path.

Example of a config file:
```
apiVersion: ignite.weave.works/v1alpha2
kind: Configuration
metadata:
  name: production-config
spec:
  runtime: containerd
  networkPlugin: cni
  sandbox:
    oci: "docker.io/darkowlzz/ignite:dev"
  vm:
    cpus: 1
    memory: "2GB"
    diskSize: "1GB"
    image:
      oci: "weaveworks/ignite-ubuntu"
    kernel:
      oci: "darkowlzz/ignite-kernel:5.3"
```
When a config file is used, ignite command execution logs the name of the config file being used:
```
# ignite run weaveworks/ignite-ubuntu:latest --ssh --name my-vm1
INFO[0000] Using ignite configuration file: /etc/ignite/config.yaml 
INFO[0001] Created VM with ID "258fc2e5ff2c1c75" and name "my-vm1" 
INFO[0001] Started Firecracker VM "258fc2e5ff2c1c75" in a container with ID "ignite-258fc2e5ff2c1c75" 
```

A local config file can be passed as:
```
# ignite run weaveworks/ignite-ubuntu --name my-vm1 --ignite-config=ignite-config.yaml
```

Related to #315


darkowlzz:
This supports a subset of the configurations discussed in #315 . Will add more options based on the feedback. The default config file path may also change based on the feedback.

cmd/ignite/run/create.go
```diff
@@ -15,99 +15,144 @@ import (
...
+	VMName        string
+	VMCPUs        uint64
+	KernelCmdLine string
+	VMMemory      meta.Size
+	VMDiskSize    meta.Size
+	// Use string for KernelOCI instead of OCIImageRef because OCIImageRef can't
+	// be created without setting a valid value. Need zero valued variable to
+	// know if the flag was set or not.
+	KernelOCI string
+	VMStorage api.VMStorageSpec
```
darkowlzz: This replaced the VM object with individual elements to be able to know if a flag is set or not. When using a VM object, the VM properties are set to the defaults. Using that as flag variables makes it hard to know if a flag was set or empty. With this, these properties are set on the base VM configuration only when the value of these flag variables are non-zero.

darkowlzz:
New implementation in #601 . Closing this.

Issue #465, opened 2019-10-01
could not parse image with a tag: image based on digest fails


lucidprogrammer:
Looks like we cannot create VM based on a specific image hash. 
`
sudo ignite launch weaveworks/ignite-kubeadm@sha256:928f55c6162a50d3278e334ac0d7a6e0bf7c765d9c6dd835b935f0ca73e398ca \
    --cpus 2 \
    --memory 1GB \
    --ssh \
    --name master-1
FATA[0000] could not parse image "weaveworks/ignite-kubeadm@sha256:928f55c6162a50d3278e334ac0d7a6e0bf7c765d9c6dd835b935f0ca73e398ca" with a tag
`
It was ideal if this feature is available

chanwit:
Thank you for reporting!

stealthybox:
`ignite launch` is not a valid subcommand, but I am able to reproduce this with `ignite run`:
```bash
sudo ignite run --log-level=trace weaveworks/ignite-kubeadm@sha256:928f55c6162a50d3278e334ac0d7a6e0bf7c765d9c6dd835b935f0ca73e398ca 
TRAC[0000] Populating providers...                      
TRAC[0000] Initializing the containerd runtime provider... 
TRAC[0000] Initializing the CNI provider...             
FATA[0000] could not parse image "weaveworks/ignite-kubeadm@sha256:928f55c6162a50d3278e334ac0d7a6e0bf7c765d9c6dd835b935f0ca73e398ca" with a tag 

```

stealthybox:
I've begun working on this.
#467 allows us to do some functional testing.

I fixed the parser on this branch:
https://github.com/stealthybox/ignite/blob/fix-oci-ref/pkg/apis/meta/v1alpha1/image.go

Just chasing down runtime issues now.

stealthybox:
`ignite image import` of digest refs works with `--runtime docker` / fails with `--runtime containerd`

Once the image is imported correctly into ignite, named by the digest, `ignite run` fails with both runtimes. `ignite-spawn` seems to time out.
Initial suspicion is that the sandbox is not inflated correctly for digest images.

A storage refactor will be necessary to clean up these bugs related to digests.

ps: I rebased my branch

stealthybox:
Sunny's patch has to do with supporting tar image import.
The metadata that results from a `docker image save` may include image name.
The result from from `docker image load` will not have a name/tag.
Overriding the name/tag is optional on `docker image import`.

PR #556, opened 2020-03-09
Fix OCIString parsing of local docker images

darkowlzz:
Local docker images don't have repo name in the OCI string.
Example: `docker://sha256:fce289e99eb9bca977dae136fbe2a82b6b7d4c372474c9235adc1741675f587e`

URL parsing OCI string of local docker image fails because of the
invalid url format.

I was working on local image import from a tar file and came across this issue when importing with runtime docker. Importing with containerd resulted in proper OCI string with repo name in it. When `/var/lib/firecracker/image` contains any image metadata with invalid OCI source ID, any ignite execution that reads images starts failing with errors like:
```
FATA[0000] v1alpha2.Image.Status: v1alpha2.ImageStatus.OCISource: v1alpha2.OCIImageSource.Size: ID: unmarshalerDecoder: parse docker://sha256:fce289e99eb9bca977dae136fbe2a82b6b7d4c372474c9235adc1741675f587e: invalid port ":fce289e99eb9bca977dae136fbe2a82b6b7d4c372474c9235adc1741675f587e" after host, error found in #10 byte of ...|1675f587e","size":"1|..., bigger context ...|977dae136fbe2a82b6b7d4c372474c9235adc1741675f587e","size":"1840B"}}}|... 
```

After debugging, the actual error was about the URL parse failure at `parseOCIString()`:
```
parse docker://sha256:fce289e99eb9bca977dae136fbe2a82b6b7d4c372474c9235adc1741675f587e: invalid port ":fce289e99eb9bca977dae136fbe2a82b6b7d4c372474c9235adc1741675f587e" after host
```
Adds tests to verify the fix.

### DIFF
--- a/pkg/apis/meta/v1alpha1/image.go
+++ b/pkg/apis/meta/v1alpha1/image.go
@@ -6,6 +6,10 @@ import (
 	"net/url"
 	"strings"
 
+	// "github.com/opencontainers/go-digest" requires us to load the algorithms
+	// that we want to use into the binary. (it calls algorithm.Available)
+	_ "crypto/sha256"
+
 	"github.com/containers/image/docker/reference"
 	"github.com/opencontainers/go-digest"
 )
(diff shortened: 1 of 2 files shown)


pkg/apis/meta/v1alpha1/image.go
```diff
@@ -6,6 +6,10 @@ import (
 	"net/url"
 	"strings"
 
+	// "github.com/opencontainers/go-digest" requires us to load the algorithms
+	// that we want to use into the binary. (it calls algorithm.Available)
+	_ "crypto/sha256"
```
darkowlzz: Got this from @stealthybox 's branch https://github.com/stealthybox/ignite/blob/fix-oci-ref/pkg/apis/meta/v1alpha1/image.go#L9-L11 . Without this, the test run fails with "unsupported digest algorithm" error.

stealthybox:
Related #465

chanwit:
LGTM

PR #557, opened 2020-03-10
[POC] Import images from tar file

darkowlzz:
Implements importing local container images from tar file. Uses a new
subcommand `tarimport` for now, since `import` command is already used
for pulling the image from image repository.
Adds a new function `ImportImage()` in the runtime interface and
implements them for docker and containerd runtimes. The images are first
imported into the runtime's image store and then copied to ignite's
backend store as ignite VM images.

Usage:
```
ignite image tarimport hello-world.tar
```

`tarimport` is a placeholder subcommand for now. It'll be changed to something better after some feedback about the implementation.

### DIFF
--- a/pkg/source/docker.go
+++ b/pkg/source/docker.go
@@ -29,6 +29,10 @@ func (ds *DockerSource) Ref() meta.OCIImageRef {
 	return ds.imageRef
 }
 
+func (ds *DockerSource) SetRef(imageRef meta.OCIImageRef) {
+	ds.imageRef = imageRef
+}
+
 func (ds *DockerSource) Parse(ociRef meta.OCIImageRef) (*api.OCIImageSource, error) {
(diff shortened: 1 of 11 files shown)


pkg/source/docker.go
```diff
@@ -29,6 +29,10 @@ func (ds *DockerSource) Ref() meta.OCIImageRef {
 	return ds.imageRef
 }
 
+func (ds *DockerSource) SetRef(imageRef meta.OCIImageRef) {
+	ds.imageRef = imageRef
+}
```
darkowlzz: `SetRef` is required to read the correct imageRef in `DockerSource.Reader` when a single image tar file contains multiple images. DockerSource importing multiple images require this option to set the imageRef of the DockerSource when the images are read.

PR #560, opened 2020-03-14
Add --require-name to create and run commands

darkowlzz:
Adds `--require-name` flag to create and run commands to fail when a VM
name is not passed. This can be used along with config file to avoid
generating different resource from the same config file due to generated
name and UID.

Adds tests to verify the change.

Follow up of https://github.com/weaveworks/ignite/pull/525

### DIFF
--- a/cmd/ignite/run/create_test.go
+++ b/cmd/ignite/run/create_test.go
@@ -1,0 +1,168 @@
+package run
+
+import (
+	"net"
+	"reflect"
+	"testing"
+
+	"github.com/weaveworks/gitops-toolkit/pkg/runtime"
+	api "github.com/weaveworks/ignite/pkg/apis/ignite"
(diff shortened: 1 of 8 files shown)


cmd/ignite/run/create_test.go
```diff
@@ -0,0 +1,204 @@
...
+			err: true,
+		},
+		{
+			name: "require-name with VM config",
+			createFlag: &CreateFlags{
+				ConfigFile:  fmt.Sprintf("testdata%c%s", filepath.Separator, "input/create-config-no-name.yaml"),
+				RequireName: true,
+			},
+			err: true,
+		},
```
darkowlzz: These cases cover only the failure cases of `NewCreateOptions`. Success case would lead to actual image imports and would require setting up a back-end storage as well. These tests are enough for checking the `--require-name` flag.

darkowlzz:
Added require-name check for VM config file as well.

PR #563, opened 2020-03-16
Update vendored libraries

luxas:
Just bumping everything except docker (we can check that later, but we should remove it anyways soon)

This is done by `rm -r vendor go.mod go.sum && make tidy-in-docker`

cc @chanwit @stealthybox


luxas:
IDK why this fails on go mod stuff... works on my machine xD
Probably safe to merge anyways, after e2e testing @stealthybox @chanwit

PR #571, opened 2020-04-01
Add OCI tool choice

bbros-dev:
Only alters kernel build scripts.
Preserves existing behavior.
Defaults to `docker` when `podman` is not available.

Tested as working via `make`

Signed-off-by: Begley Brothers Inc <begleybrothers@gmail.com>

### DIFF
--- a/images/kernel/upgrade-config.sh
+++ b/images/kernel/upgrade-config.sh
@@ -21,7 +21,13 @@ if [[ ${FROM} != ${TO} ]]; then
     cp ${FROM} ${TO}
 fi
 
-docker run -it \
+if ! [ -x "$(command -v podman)" ]; then
+  oci_tool=podman
+else
+  oci_tool=docker
+fi
+
+${oci_tool} run -it \
     ${ARCH_PARAMETER} \
 	-v $(pwd)/${TO}:/tmp/.config \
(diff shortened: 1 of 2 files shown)


images/kernel/upgrade-config.sh
```diff
@@ -21,7 +21,13 @@ if [[ ${FROM} != ${TO} ]]; then
     cp ${FROM} ${TO}
 fi
 
-docker run -it \
+if ! [ -x "$(command -v podman)" ]; then
+  oci_tool=podman
+else
+  oci_tool=docker
```
stealthybox, line 27: Thanks for the patch :wave:

Would `DOCKER=podman make build` be sufficient for you?
This would be consistent with the main `Makefile` we already have.

Other usage would be:
`DOCKER="sudo docker" make build`

The kernel build helper scripts would also need to be modified to use this ENV var to override docker.
We could do a followup PR if you don't have the bandwidth to do so.

bbros-dev: No worries: `DOCKER=podman`.  Will try get to that this week and force update this.

stealthybox: thanks!

PR #578, opened 2020-04-10
Update kernel configs

luxas:
I got the new Firecracker reference configs from here: https://github.com/firecracker-microvm/firecracker/tree/master/resources

WARNING: We should make sure these still work with firekube cc @chanwit 
The FC team removed a lot of things (streamlined the kernel), and we might depend on some of them. If so, we should add patches we need to `images/kernel/config-patches`.

cc @stealthybox let's use these kernels in the next v0.7.0 release


gaby:
@stealthybox This PR can be close now

PR #580, opened 2020-04-11
Unify ssh and exec commands

darkowlzz:
This change combines ssh and exec commands to use the same code. ssh no longer
uses ssh binary on the host but uses go's crypto/ssh package to create a
terminal session.

Moves `newSignerForKey`, `newSSHConfig` and joinShellCommand functions from
exec.go to ssh.go, keeping all ssh related functions in the same file.


darkowlzz:
e2e test seem to be flaky. It ran fine on my fork https://travis-ci.com/github/darkowlzz/ignite/builds/159927085

stealthybox:
Please rebase on master which now includes #581 :)

darkowlzz:
Added a defer with `os.Exit` after a ssh connection is establish. Since the function will no longer be able to return any error, any error that's found after this are printed so that it's visible to the user. The exit code is set based on the exit code from the ssh session.

PR #592, opened 2020-05-04
Enable unit tests

darkowlzz:
Fix a few tests and skip e2e tests and incomplete package pkg/dm from unit test make target.
Separates the tests that require root, like the containerd client test.
Non-root tests can be run with `make test` and root tests can be run with `make root-test`.
Run the tests in TravisCI.


pkg/client/client_test.go
```diff
@@ -1,25 +1,32 @@
...
+// )
 
-	t.Error(c.VMs().Delete("1234"))
+// var c = NewClient(cache.NewCache(
+// 	storage.NewGenericStorage(
+// 		storage.NewGenericRawStorage("/tmp/foo"), scheme.Serializer)))
 
-	vms, err := c.VMs().List()
-	t.Error(len(vms), err)
-}
```
darkowlzz, line 17: This and the following tests try to get a VM without creating the VM, which results in failure. Should this be removed for other tests that already use the client properly? Or should the test be fixed by creating a new VM object, querying and deleting it?

.circleci/config.yml
```diff
@@ -24,8 +24,13 @@ jobs:
...
     - run: sudo mkdir -p /go/bin && sudo chown -R circleci /go/bin
-    # Run unit tests, TODO
-    #- run: go test ./...
+    # Run unit tests.
+    - run:
+        name: unit tests
+        command: make unit-test-in-docker
+        # command: $GO test -mod=vendor $($GO list ./... | grep -v /e2e | grep -v /pkg/dm | grep -v pkg/runtime/containerd)
+        # environment:
+        #   - GO: "sudo /usr/local/go/bin/go"
```
darkowlzz, line 33: Tried running the unit tests on host but `go list` didn't seem to work with the go on the host. https://circleci.com/gh/weaveworks/ignite/1342?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

If running go test on host directly with sudo works, containerd can be installed to run the containerd client tests. This is excluded when running in container.
We can also try running containerd in a test container and run the test against it.

pkg/apis/meta/v1alpha1/image_test.go
```diff
@@ -26,11 +26,13 @@ func TestNewOCIImageRef(t *testing.T) {
...
 		actual, err := NewOCIImageRef(rt.in)
 		if (err != nil) != rt.err {
-			t.Errorf("expected error %t, actual: %v", rt.err, err)
+			t.Fatalf("expected error %t, actual: %v", rt.err, err)
 		}
-
-		if actual.String() != rt.out {
-			t.Errorf("expected %q, actual: %q", rt.out, actual.String())
+		// Check actual string when there's no error.
+		if err == nil {
```
stealthybox, line 32: +1

.circleci/config.yml
darkowlzz: Moved the tests to TravisCI that has containerd installed for e2e tests.

pkg/client/client_test.go
darkowlzz: Removed these tests. They seemed to be for development purpose only.

PR #598, opened 2020-05-16
Add sandbox-image VM flag

darkowlzz:
This adds a new flag `--sandbox-image` for setting the VM sandbox image.
The VM spec contains a "sandbox" field, similar to "image" and "kernel"
fields.

Adds a new constant `DEFAULT_SANDBOX_IMAGE_NAME`.

Adds unit test for sandbox-image flag.

Autogen updates the deps. Deps update are in a separate commit.


darkowlzz:
The CI fails because [`version.ImageTag()`](https://github.com/weaveworks/ignite/blob/fe4b2aa7cf4c9f4fd1591f15f4f3a1c6b8cb006c/pkg/version/version.go#L42) is not able to return a proper version for the sandbox image tag, resulting in error when running ignite due to missing image tag.

```console
panic: interface conversion: interface is nil, not reference.NamedTagged [recovered]

	panic: interface conversion: interface is nil, not reference.NamedTagged
```

This maybe due to the shallow clone of the repo in the CI.

Building manually with full git history works properly and the sandbox image tag is set as expected:
```console
$ sudo bin/ignite create -h
...
Flags:
...
      --sandbox-image oci-image   Specify an OCI image for the VM sandbox (default weaveworks/ignite:v0.6.0-262-fe4b2aa7cf4c9f)
...
```

`weaveworks/ignite:v0.6.0-262-fe4b2aa7cf4c9f` in the above output.

We may need to improve how `version.ImageTag()` works.

stealthybox:
This ended up not being the specific reason the tests were failing:
```
The CI fails because version.ImageTag() is not able to return a proper version for the sandbox image tag, resulting in error when running ignite due to missing image tag.
```

I was able to modify the code in the first commit I added make the image defaulting logic work without the help of ldflags and gitVersion.

The tests were still failing locally in my git copy.
The failures were due to the lack of missing fields on synthetic VM objects created within the tests.
These objects are not created using the defaulting logic from api-machinery, so their OCI fields need to manually be set.
I did not fix the lack of defaulting functions, but I was able to manually add static values for the missing fields in the second commit, and the tests are now passing.

stealthybox:
I was able to test this field also properly defaults and can be overridden using `sudo bin/ignite run --config ../ignite-scratch/anon-vm.json`

darkowlzz:
This is great. A lot of improvements 🎉 Thanks a lot.

Issue #355, opened 2019-08-19
Remove v1alpha1 in v0.8.0


luxas:
We will support upgrades so that you can seamlessly go one minor version up at any time without breaking, but we won't hold an old API version left in the binary more releases than necessary.

Upgrading from v1alpha1 (the only option in v0.4.0) to v1alpha2 was possible in v0.5.0, but is not needed in v0.6.0 anymore as everything touched by v0.5.0 is always at v1alpha2.

**Update:** We'll remove `v1alpha1` from v0.7.0, as that most likely introduces `v1alpha3`. Maintaining three API versions in parallel would take up too much time/effort. v0.6.0 will thus have `v1alpha1` still available (use discouraged).

luxas:
@chanwit

chanwit:
got it!

luxas:
@darkowlzz could you please do this? when this is done, we can create a new `v1alpha3`, to which you can add the Config object, ref: #601

darkowlzz:
@luxas sure. Is it just removing the package `pkg/apis/ignite/v1alpha1` and updating any code that uses it to `v1alpha2`?

luxas:
Yes. After that we can duplicate v1alpha2 into v1alpha3, and then start modifying v1alpha3 to have the new things we want.

PR #601, opened 2020-05-31
Add support for ignite configuration file

darkowlzz:
This adds a new API type `Configuration` which is used to define the
default configurations of ignite. A Configuration object can be defined
in a config file as:

```yaml
apiVersion: ignite.weave.works/v1alpha3
kind: Configuration
metadata:
  name: test-config
spec:
  runtime: containerd
  networkPlugin: cni
  vmDefaults:
    memory: "2GB"
    diskSize: "3GB"
    cpus: 2
    sandbox:
      oci: weaveworks/ignite:dev
    kernel:
      oci: weaveworks/ignite-kernel:4.19.47
    ssh: true
```

By default, ignite looks for the config file at /etc/ignite/config.yaml.
If this file is found, ignite uses it to configure the ignite defaults.
`--ignite-config` flag can be used to explicitly pass a configuration
file.

- Changes the flag override behavior when using a VM config file. The flags are no longer ignored and override the VM configuration.
- Debug log for the configuration used:
Default config:
```console
$ sudo bin/ignite run weaveworks/ignite-ubuntu --name my-vm --ssh --log-level=debug
DEBU[0000] Using ignite default configurations
```
With global config:
```console
$ sudo bin/ignite run weaveworks/ignite-ubuntu --name my-vm --ssh --log-level=debug
DEBU[0000] Found default ignite configuration file /etc/ignite/config.yaml
DEBU[0000] Using ignite configuration file /etc/ignite/config.yaml
```
- Adds e2e tests to verify the behavior.
- Adds new document `docs/ignite-configuration.md` with examples of the ignite configuration.

Related to #315


luxas:
I'd add this to a new v1alpha3 API version, in order to not break v1alpha2 compatibility. ref: https://github.com/weaveworks/ignite/issues/355

darkowlzz:
Rebased on v1alpha3 API.

cmd/ignite/cmd/root.go
```diff
@@ -115,9 +153,27 @@ func addGlobalFlags(fs *pflag.FlagSet) {
...
 // AddQuietFlag adds the quiet flag to a flagset
 func AddQuietFlag(fs *pflag.FlagSet) {
 	fs.BoolVarP(&logs.Quiet, "quiet", "q", logs.Quiet, "The quiet mode allows for machine-parsable output by printing only IDs")
 }
+
+// getConfigFromFile reads a config file and returns ignite configuration.
+func getConfigFromFile(configPath string) (*api.Configuration, error) {
+	componentConfig := &api.Configuration{}
+
+	if err := scheme.Serializer.DecodeFileInto(configPath, componentConfig); err != nil {
```
twelho, line 168: TODO: Fix libgitops `DecodeFileInto` to not allow empty files

pkg/apis/ignite/v1alpha3/types.go
```diff
@@ -237,3 +239,19 @@ type VMStatus struct {
...
+	runtime.ObjectMeta `json:"metadata"`
+
+	Spec ConfigurationSpec `json:"spec"`
+}
+
+// ConfigurationSpec defines the ignite configuration.
+type ConfigurationSpec struct {
+	Runtime       igniteRuntime.Name       `json:"runtime,omitempty"`
+	NetworkPlugin igniteNetwork.PluginName `json:"networkPlugin,omitempty"`
+	VM            VMSpec                   `json:"vm,omitempty"`
```
twelho, line 256: Rename to VMDefaults ... `vmDefaults`

cmd/ignite/cmd/root.go
```diff
@@ -52,6 +58,38 @@ func NewIgniteCommand(in io.Reader, out, err io.Writer) *cobra.Command {
...
+			} else {
+				// Check the default config location.
+				if _, err := os.Stat(constants.IGNITE_CONFIG_FILE); !os.IsNotExist(err) {
+					log.Infof("Found default ignite configuration file %s", constants.IGNITE_CONFIG_FILE)
+					configFilePath = constants.IGNITE_CONFIG_FILE
+				}
+			}
+
+			if configFilePath != "" {
+				log.Infof("Using ignite configuration file %s", configFilePath)
```
stealthybox, line 77: We decided to keep these logs:
- increase the log level so it's not noisy in Info
- also log when there is no config found at all

cmd/ignite/cmd/root.go
```diff
@@ -52,6 +58,38 @@ func NewIgniteCommand(in io.Reader, out, err io.Writer) *cobra.Command {
...
+				configFilePath = configPath
+			} else {
+				// Check the default config location.
+				if _, err := os.Stat(constants.IGNITE_CONFIG_FILE); !os.IsNotExist(err) {
+					log.Infof("Found default ignite configuration file %s", constants.IGNITE_CONFIG_FILE)
+					configFilePath = constants.IGNITE_CONFIG_FILE
+				}
+			}
+
+			if configFilePath != "" {
```
twelho, line 76: Also handle the config for `ignited`?

e2e/config_test.go
```diff
@@ -0,0 +1,234 @@
...
+				runArgs = append(runArgs, "--config="+vmConfigFileName)
+			}
+
+			// Append the args to the run args for override flags.
+			runArgs = append(runArgs, rt.args...)
+			runCmd := exec.Command(
+				igniteBin,
+				runArgs...,
+			)
+			runOut, _ := runCmd.CombinedOutput()
```
twelho, line 192: Catch exit code using the returned error

cmd/ignite/cmd/root.go
darkowlzz: Maybe in a follow-up PR?

cmd/ignite/cmd/root.go
```diff
@@ -115,9 +153,27 @@ func addGlobalFlags(fs *pflag.FlagSet) {
 	logflag.LogLevelFlagVar(fs, &logLevel)
 	runtimeflag.RuntimeVar(fs, &providers.RuntimeName)
 	networkflag.NetworkPluginVar(fs, &providers.NetworkPluginName)
+	fs.StringVar(&configPath, "ignite-config", "", "Ignite configuration path")
```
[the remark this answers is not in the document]
darkowlzz: I've added a document about this in `docs/ignite-configuration.md`. I hope this will make it more clear.
Should this still be more descriptive?

cmd/ignite/run/create_test.go
```diff
@@ -120,53 +251,28 @@ func TestConstructVMFromCLI(t *testing.T) {
...
-			name: "with no VM name and --require-name flag set",
-			createFlag: &CreateFlags{
-				VM:          &api.VM{},
-				RequireName: true,
-			},
-			args: []string{testImage},
-			err:  true,
-		},
-		{
-			name: "with sandbox image",
```
twelho, line 139: This is already tested in e2e so it's fine to exclude it here (tricky to implement).

cmd/ignite/run/create_test.go
```diff
@@ -120,53 +251,28 @@ func TestConstructVMFromCLI(t *testing.T) {
...
 				},
 			},
-			args: []string{testImage},
 			wantSSH: api.SSH{
 				Generate:  true,
 				PublicKey: "some-pub-key",
 			},
 		},
-		{
-			name: "with no VM name and --require-name flag set",
```
twelho, line 130: This was tested twice, check https://github.com/weaveworks/ignite/pull/601/files#diff-d0bf10b7da8bd0e055593597d507f004R306

cmd/ignite/run/create_test.go
```diff
@@ -120,53 +251,28 @@ func TestConstructVMFromCLI(t *testing.T) {
...
-			args:             []string{testImage},
-			wantSandboxImage: sandboxOCIRef,
-		},
 	}
 
 	for _, rt := range tests {
 		t.Run(rt.name, func(t *testing.T) {
-			err := rt.createFlag.constructVMFromCLI(rt.args)
+			vm := rt.createFlag.VM
+			fs := flag.NewFlagSet("test", flag.ExitOnError)
```
twelho, line 264: This does not test flag validity since that would need the flag set to be populated (needs additional logic), but the flag handling will need to be refactored at some point anyway so this is fine.

cmd/ignite/run/create_test.go
```diff
@@ -221,7 +322,23 @@ func TestNewCreateOptions(t *testing.T) {
 
 	for _, rt := range tests {
 		t.Run(rt.name, func(t *testing.T) {
-			_, err := rt.createFlag.NewCreateOptions([]string{})
+			// Setup storage backend.
+			dir, err := ioutil.TempDir("", "ignite")
+			if err != nil {
+				t.Fatalf("failed to create storage for ignite: %v", err)
+			}
+			defer os.RemoveAll(dir)
```
twelho, line 330: Catch this error with `util.DeferErr` (will be done in a separate PR)

cmd/ignite/run/create_test.go
```diff
@@ -221,7 +322,23 @@ func TestNewCreateOptions(t *testing.T) {
...
+
+			storage := cache.NewCache(
```
twelho, line 332: This should use Ignite's configured providers to have consistency between the tests and built binary, can be done in a separate PR.

PR #602, opened 2020-06-01
Update OS and kernel versions

luxas:
@darkowlzz @stealthybox 

```
make -C images build-all
make -C images/kernel upgrade
make -C images/kernel build
```


casibbald:
Any chance we can get a 5.6 kernel on Ubuntu 20.04?

luxas:
> Any chance we can get a 5.6 kernel on Ubuntu 20.04?

Should be (pretty easily) doable.

stealthybox:
@casibbald that's a very recent kernel version -- I'm curious what your use-case is :)

My concern if we were to publish latest kernels would be not being able to update them frequently without CI running on cron.

I think we're happy to publish the newer kernel build even if it's just a one-time/infrequent thing though.

stealthybox:
Also noting I was wrong about the kubernetes version
Latest is `1.18.3`

PR #604, opened 2020-06-02
Bump kernel configurations

luxas:
Just re-ran `make -C images/kernel upgrade` for new configs


luxas:
Self-merging this as it's pretty much just a autogenerated follow up of #602

PR #607, opened 2020-06-02
Unify some things around SSH handling

luxas:
@stealthybox @darkowlzz

### DIFF
--- a/cmd/ignite/cmd/cmdutil/flags.go
+++ b/cmd/ignite/cmd/cmdutil/flags.go
@@ -2,6 +2,7 @@ package cmdutil
 
 import (
 	"github.com/spf13/pflag"
+	"github.com/weaveworks/ignite/pkg/constants"
 )
 
 // This file contains a collection of common flag adders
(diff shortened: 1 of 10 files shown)


cmd/ignite/cmd/cmdutil/flags.go
```diff
@@ -20,3 +21,8 @@ func AddInteractiveFlag(fs *pflag.FlagSet, interactive *bool) {
 func AddForceFlag(fs *pflag.FlagSet, force *bool) {
 	fs.BoolVarP(force, "force", "f", *force, "Force this operation. Warning, use of this mode may have unintended consequences.")
 }
+
+func AddSSHFlags(fs *pflag.FlagSet, identityFile *string, timeout *uint32) {
+	fs.StringVarP(identityFile, "identity", "i", "", "Override the vm's default identity file")
+	fs.Uint32Var(timeout, "timeout", constants.SSH_DEFAULT_TIMEOUT_SECONDS, "Timeout waiting for connection in seconds")
```
stealthybox, line 27: Should this be `--ssh-timeout`?
Since this is a VM creation flag, there could be other timeouts

cmd/ignite/run/start.go
```diff
@@ -74,12 +76,16 @@ func dialSuccess(vm *ignite.VM, seconds int) error {
 	for i := 0; i < seconds*perSecond; i++ {
 		conn, dialErr := net.DialTimeout("tcp", addr, delay)
 		if conn != nil {
-			defer conn.Close()
+			conn.Close()
```
stealthybox, line 79: From group review:
This fixes a bug where the connections would stay open as this loop continued.
The defers stack at the end of the function which is not within the loop.

PR #609, opened 2020-06-02
A bit more details on the ARM 64-bit documentation

luxas:
@stealthybox @darkowlzz

### DIFF
--- a/docs/arm.md
+++ b/docs/arm.md
@@ -1,53 +1,47 @@
-# Using ignite on ARM
+# Using Ignite on ARM64
 
-Ignite had been tested to function on Raspberry Pi 3 and 4.
-These instructions should be helpful in using ignite on other ARM machines such as those in the cloud.
+Ignite has been tested to function on Raspberry Pi 4, as well as arm64 cloud machines.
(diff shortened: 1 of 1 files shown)


docs/arm.md
```diff
@@ -1,53 +1,47 @@
...
-```
+## containerd
 
-Now you may install these binaries into the appropriate place on your `$PATH` and configure your containerd [`config.toml`](https://github.com/containerd/containerd/blob/master/docs/man/containerd-config.toml.5.md) and [service unit/init](https://github.com/containerd/containerd/blob/master/containerd.service) file.
+For Raspberry Pi OS, you can follow the normal installation procedure as for Ubuntu.
 
-# Docker
-If you choose to use the docker runtime instead of ignite's default support for containerd, you may install Docker's arm64 binaries.
-In our test setup, we installed docker with `emerge`.
+We don't know of a built-in way to install containerd on Gentoo, but you can extract the deb package manually from [Ubuntu's arm64 package](http://ports.ubuntu.com/ubuntu-ports/pool/main/c/containerd/).
```
darkowlzz, line 27: This is much better 👍

PR #610, opened 2020-06-02
Bump default kernel to 4.19.125

stealthybox:
cc @darkowlzz @luxas


stealthybox:
Note CI failure is because kernels are staged but not published

PR #615, opened 2020-06-09
Use the new libgitops GitDirectory implementation

luxas:
ref: https://github.com/weaveworks/libgitops/pull/8 https://github.com/weaveworks/libgitops/issues/4

cc @twelho


pkg/gitops/gitops.go
```diff
@@ -1,24 +1,24 @@
...
-	log.Infof("Whenever changes are pushed to the %s branch, Ignite will apply the desired state locally\n", branch)
+	log.Info("Whenever changes are pushed to the target branch, Ignite will apply the desired state locally\n")
 
 	// Construct the GitDirectory implementation which backs the storage
-	gitDir := gitdir.NewGitDirectory(url, branch, paths, syncInterval)
+	gitDir, err := gitdir.NewGitDirectory(url, opts)
+	if err != nil {
+		return err
+	}
+	// TODO: Run gitDir.Cleanup() on SIGINT
```
twelho, line 21: Related implementation: https://github.com/weaveworks/ignite/blob/d98868084332496c56813350665e01cf540091a8/pkg/container/firecracker.go#L147

luxas: Yep, I know, and we have one already in ignited, but in an other place. For laters

PR #617, opened 2020-06-11
Use the new knownhosts package in fluxcd/toolkit

luxas:
ref: https://github.com/weaveworks/libgitops/pull/9

/cc @twelho @stealthybox


cmd/ignited/cmd/gitops.go
```diff
@@ -85,15 +106,25 @@ func addGitOpsFlags(fs *pflag.FlagSet, f *gitOpsFlags) {
...
+	if !strings.Contains(path, "~") {
+		return path, nil
+	}
+	// get the HOME env var and fail if not found
+	homeEnv := os.Getenv("HOME")
+	if len(homeEnv) == 0 {
+		return "", fmt.Errorf("Couldn't lookup home directory for user")
+	}
+	// replace the first tilde in the path with $HOME
+	return strings.Replace(path, "~", homeEnv, 1), nil
```
twelho, line 129: This will perform an illicit replace if the ~ is not the first character of the string. We should likely do something like this (which also retrieves the home directory correctly): https://stackoverflow.com/a/43578461 if this function is really needed (see comment above)

cmd/ignited/cmd/gitops.go
```diff
@@ -61,9 +67,24 @@ func NewCmdGitOps(out io.Writer) *cobra.Command {
 			}
 			if f.identityFile != "" {
 				var err error
+				// support ~ prefixes in the path
```
twelho, line 70: Have you verified that this is necessary, since the shell will expand ~ before it is passed to the program? Unless this this config is read from a file, I don't see the need to do manual path expansion.

luxas: it is needed, as we have the (non-expanded) part as a default

cmd/ignited/cmd/gitops.go
luxas: true. using https://github.com/mitchellh/go-homedir/blob/master/homedir.go now :+1:

Issue #618, opened 2020-06-15
Link checking of ./site hits too many Github URLs


dholbach:
Merging #573 (https://github.com/weaveworks/ignite/commit/aedb9cd47f4f1c2717befc8fac3a243ce83edc2f) link checking failed (https://github.com/weaveworks/ignite/runs/771426681) because of:

```console
site/api/ignite_v1alpha1/index.html
	ERROR	https://github.com/weaveworks/ignite/tree/master/pkg/apis/ignite/v1alpha1/types.go?s=7766:7861#L207
		Too Many Requests (HTTP error 429)
```

We simply hit too many Github URLs.

Here is the regex for links we ignore:

```
-x "(github.com/weaveworks/ignite($|/issue|/pull|/releases|/commit/|/releases)|squidfunk.github.io|ignite.readthedocs.io)"
```

I feel URLs outside of Ignite's Github should be checked, so we should potentially look at checking some of Ignite's links a bit less aggressively. 

@luxas Are the links pointing to `https://github.com/weaveworks/ignite/tree/master/pkg/apis/ignite/v1alpha1/*` not all auto-generated? Can we exclude those?

Currently there are:

```console
[daniel@reef ignite ]$ grep -orE "\"(https://github.com/weaveworks/ignite\S*?)\"" site/ | cut -d':' -f2-3 | sort -u
"https://github.com/weaveworks/ignite"
"https://github.com/weaveworks/ignite/blob/master/CHANGELOG.md#v040"
"https://github.com/weaveworks/ignite/blob/master/CHANGELOG.md#v041"
"https://github.com/weaveworks/ignite/blob/master/docs/architecture.png"
"https://github.com/weaveworks/ignite/blob/master/docs/declarative-config.md"
"https://github.com/weaveworks/ignite/blob/master/docs/dependencies.svg"
"https://github.com/weaveworks/ignite/blob/master/docs/installation.md"
"https://github.com/weaveworks/ignite/blob/master/docs/prometheus.md"
"https://github.com/weaveworks/ignite/blob/master/images/alpine/Dockerfile"
"https://github.com/weaveworks/ignite/blob/master/images/amazon-kernel/Dockerfile"
"https://github.com/weaveworks/ignite/blob/master/images/amazonlinux/Dockerfile"
"https://github.com/weaveworks/ignite/blob/master/images/centos/Dockerfile"
"https://github.com/weaveworks/ignite/blob/master/images/kernel/Dockerfile"
[...]

PR #619, opened 2020-06-15
link-checking: Ignore /pkg/apis links, auto-generated docs

dholbach:
- Update theme version as well.

Closes: #618

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -45,7 +45,7 @@ jobs:
         id: lc2
         uses: peter-evans/link-checker@v1
         with:
-          args: -d site/ -r site/ -x "(github.com/weaveworks/ignite($|/issue|/pull|/releases|/commit/|/releases)|squidfunk.github.io|ignite.readthedocs.io)"
+          args: -d site/ -r site/ -x "(github.com/weaveworks/ignite($|/issue|/pull|/releases|/commit/|/releases|/tree/master/pkg/apis)|squidfunk.github.io|ignite.readthedocs.io)"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/check-links.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
         id: lc2
         uses: peter-evans/link-checker@v1
         with:
-          args: -d site/ -r site/ -x "(github.com/weaveworks/ignite($|/issue|/pull|/releases|/commit/|/releases)|squidfunk.github.io|ignite.readthedocs.io)"
+          args: -d site/ -r site/ -x "(github.com/weaveworks/ignite($|/issue|/pull|/releases|/commit/|/releases|/tree/master/pkg/apis)|squidfunk.github.io|ignite.readthedocs.io)"
```
stealthybox, line 48: Checking `/tree/master` actually caught a valid error in the docs.
See commit: c53a6a4

Fails on master because we did not remove v1alpha1 API docs when we deprecated the API

Solution to the valid link-checking error was PR: #625

dholbach: So should I close this PR? The theme update is not strictly necessary.

darkowlzz: @dholbach hi, we had a discussion about this in the weekly meeting. We agreed that it's useful to run the link checks on the master branch only and we should disable the checks on the PR.
Can you update this PR to keep the theme update and update the github workflow for check-links to not run on pull requests?

stealthybox: Merged #656

PR #620, opened 2020-06-15
Remove ignite v1alpha1 API

### LINKED ISSUE #355: Remove v1alpha1 in v0.8.0

darkowlzz:
Remove `pkg/apis/ignite/v1alpha1/` and update the generated files with
`make autogen`.

Fixes #355


.circleci/config.yml
```diff
@@ -39,9 +39,9 @@ jobs:
     - restore_cache:
         key: bin-cache
     # Run autogeneration and see if there are differences
-    # TODO: Fix this; somehow the docs/api/ignite_v1alpha1.md ends up being malformatted in CI; it adds
+    # TODO: Fix this; somehow the docs/api/ignite_v1alpha2.md ends up being malformatted in CI; it adds
```
twelho, line 42: Could you verify if this workaround is still necessary?

darkowlzz: Removed it and looks like we no longer need it 😃 
Thanks.

Issue #633, opened 2020-07-03
`ignite-spawn` container has multiple interfaces with same MAC, networking (randomly) broken


twelho:
We've now identified the issue, the `ignite-spawn` container has multiple interfaces used for bridging sharing the same MAC address. The cause of this is still unknown, a quick fix is to (unsuccessfully) ping the __previous__ IP address given a VM by CNI and after that ping the __current__ IP address. This will open communications both ways (now you can ping the VM and the VM can access the outside world).

This has also been seen rarely with the `docker-bridge` backend where the VMs have no access to the outside world.

cc @stealthybox, @chanwit

_Originally posted by @twelho in https://github.com/weaveworks/ignite/issues/616#issuecomment-653605319_

twelho:
Duplicate MAC address issue figured out (does not fix networking on its own). We're calling `netlink.LinkSetMaster` here (and again some lines below that): https://github.com/weaveworks/ignite/blob/52a0fce1789d454d31af3862a8c918616582e4f7/pkg/container/network.go#L135

This is to bind the container-given network interface as well as the virtual VM interface to the bridge. If `netlink.LinkSetMaster` is called on a "dangling" bridge (nothing attached to it) and a virtual (TUNTAP) device, both devices will change their MAC addresses to something arbitrary, most often both get the same MAC. This does not happen if a "real" adapter device (such as the ethernet card on my PC) is attached to the bridge __first__ using that same function call. If both devices are virtual (like in Ignite's case), this might also only affect the virtual first device attached to the bridge. Anyways this is very likely a bug in `netlink`, though I'll investigate a little further before opening an issue since even when manually persisting the MAC addresses, networking still doesn't work out of the box with Flannel and `dmesg` shows the following messages:
```
br_eth0: received packet on eth0 with own address as source address (addr:a6:e4:3d:ce:c4:34, vlan:0)
```

twelho:
[...]

PR #634, opened 2020-07-03
Possible fix for the CNI problem?

luxas:
Maybe fixes: #633 
cc @twelho


luxas:
cc @chanwit @stealthybox on the above CNI issue ^

PR #647, opened 2020-07-19
Allow provider override when using config file

darkowlzz:
When using ignite configuration file, the global flags `--runtime` and
`--network-plugin` were ignored. This change enables the overrides to work.

The default runtime and network-plugin name assignment is moved from
providers into `NewIgniteCommand` and the defaults are set only when the
values are empty after applying config files and flag overrides.


darkowlzz:
Writing a test for this will be easier if we have runtime and network-plugin info in the VM object.
I've manually tested this and it doesn't change any of the existing behaviors as per the e2e tests.

PR #653, opened 2020-08-04
Fix in-container device nodes for containerd runtime

### LINKED ISSUE #646: Mounting volume device using the `--volumes` flag not working
Hello Weaveworks team,

I am trying to start a VM that uses the `--volume` flag to mount a disk device inside a VM, but all attempts to make this work are futile. Here is what I do:

```
# Create loop device
dd if=/dev/zero of=/test_vol/diskimage bs=1M count=1024
# Make it a ext4 FS
mkfs.ext4 /test_vol/diskimage
# Set up the loop device
sudo losetup /dev/loop17 /test_vol/diskimage
# Start Ignite with the --volume flag (the loop device is not mounted locally, my understanding is that Ignite / Firecracker will mount it inside the VM)
sudo ignite \
        run weaveworks/ignite-kubeadm:latest \
        --network-plugin cni \
        --cpus 2 \
        --memory 2GB \
        --ssh \
        --volumes /dev/loop17:/dev/loop1 \
        --name master-1
```

Here is the output of the command:

```
INFO[0000] Created VM with ID "e2c782438e20f3f3" and name "master-1" 
INFO[0001] Started Firecracker VM "e2c782438e20f3f3" in a container with ID "ignite-e2c782438e20f3f3" 
INFO[0002] Waiting for the ssh daemon within the VM to start... 
INFO[0013] Waiting for the ssh daemon within the VM to start... 
INFO[0023] Waiting for the ssh daemon within the VM to start... 
FATA[0033] dial tcp 10.50.59.6:22: connect: connection refused
```

There is no documentation on how to use the `--volumes` flag, I did try all sorts of variations to make this work, but nothing helps. Can someone help me out on this please?

BR

stealthybox:
Devices are not namespaced in the Linux kernel.
Device-node paths are not dependent on any particular path in a mount namespace.
They are referred to globally by their major and minor number.

This was only working before for devices that had the same path in-container as on-host.
This patch creates a device-node in the sandbox container for the proper in-container specified path.

Fixes #646

I verified this makes volumes mount properly in vm's with containerd as the runtime.

### DIFF
--- a/e2e/run_volume_test.go
+++ b/e2e/run_volume_test.go
@@ -172,6 +172,8 @@ func runVolume(t *testing.T, vmName, runtime, networkPlugin string) {
 }
 
 func TestVolumeWithDockerAndDockerBridge(t *testing.T) {
+	// TODO: https://github.com/weaveworks/ignite/issues/658
+	t.Skip("SKIPPING\nThis test fails to stop the VM within docker\nTODO: https://github.com/weaveworks/ignite/issues/658")
 	runVolume(
 		t,
 		"e2e_test_volume_docker_and_docker_bridge",
--- a/e2e/run_volume_test.go
+++ b/e2e/run_volume_test.go
@@ -1,0 +1,199 @@
+//
+// This is the e2e package to run tests for Ignite
+// Currently, we do local e2e tests only
+// we have to wait until the CI setup to allow Ignite to run with sudo and in a KVM environment.
+//
+// How to run tests:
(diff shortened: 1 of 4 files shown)


e2e/run_volume_test.go
```diff
@@ -0,0 +1,199 @@
+//
+// This is the e2e package to run tests for Ignite
+// Currently, we do local e2e tests only
+// we have to wait until the CI setup to allow Ignite to run with sudo and in a KVM environment.
```
darkowlzz, line 4: We can remove these comments now.

stealthybox:
🚢

Issue #504, opened 2019-12-14
How to re-run stopped VMs?


shinebayar-g:
1. Why `reboot` VMs doesn't work?
2. How can I re-start stopped VMs?
```
# sudo ignite vm start my-vm
FATA[0000] failed to start container for VM "989923904ae4eb43": task must be stopped before deletion: running: failed precondition
```

stealthybox:
Thanks for reporting.
This is probably a bug.
We should write an e2e test and see when it was introduced.

kobayashi:
Which ignite version are you using? For my environment, stop/start tasks are working fine.

```
$ sudo ignite run weaveworks/ignite-ubuntu --ssh --name my-vm
INFO[0001] Created VM with ID "ba6509c56933aa45" and name "my-vm" 
INFO[0001] Networking is handled by "cni"               
INFO[0001] Started Firecracker VM "ba6509c56933aa45" in a container with ID "ignite-ba6509c56933aa45" 
$ sudo ignite stop my-vm
INFO[0000] Removing the container with ID "ignite-ba6509c56933aa45" from the "cni" network 
INFO[0000] Stopped VM with name "my-vm" and ID "ba6509c56933aa45" 
$ sudo ignite start my-vm
INFO[0000] Networking is handled by "cni"               
INFO[0000] Started Firecracker VM "ba6509c56933aa45" in a container with ID "ignite-ba6509c56933aa45" 
$ sudo ignite vm ps
VM ID			IMAGE				KERNEL					SIZE	CPUS	MEMORYCREATED	STATUS	IPS		PORTS	NAME
ba6509c56933aa45	weaveworks/ignite-ubuntu:latest	weaveworks/ignite-kernel:4.19.47	4.0 GB	1	512.0 MB	23s ago	Up 8s	10.61.0.32		my-vm
$ sudo ignite version
Ignite version: version.Info{Major:"0", Minor:"6+", GitVersion:"v0.6.0-235+4b20e4f5cd8c58-dirty-dirty", GitCommit:"4b20e4f5cd8c582daf7fb8cbc0359d3ccab6b5bb", GitTreeState:"dirty", BuildDate:"2020-04-20T20:29:28Z", GoVersion:"go1.14.2", Compiler:"gc", Platform:"linux/amd64"}
Firecracker version: v0.21.1
Runtime: containerd
```

stealthybox:
Thanks for confirming on recent versions of ignite @kobayashi.

[...]

PR #661, opened 2020-08-15
Support removing networking from stopped VM

### LINKED ISSUE #660: Failed to restart a stopped VM caused by `a handled signal`
When a VM stopped by a handled signal, `ignite-spawn` will update the VM running status to `false` but it does not handle the VM networking correctly to cause the VM unable to start due to `duplicated IP allocation error`. 

## Steps to reproduce:
- `ignite run` starts a VM
- `kill -2 <ignite spawn/VM PID>`
- `ignite stop <VM ID>` failed to forcibly stop the VM
```console
➜  ignite stop b5d26a20c03fc678
FATA[0000] VM "b5d26a20c03fc678" is not running 
```
- `ignite start <VM ID>` failed with `duplicated IP allocation error`
```console
➜  ignite start b5d26a20c03fc678
ERRO[0000] failed to setup network for namespace "ignite-b5d26a20c03fc678": failed to allocate for range 0: 10.61.0.42 has been allocated to ignite-b5d26a20c03fc678, duplicate allocation is not allowed 
FATA[0000] failed to allocate for range 0: 10.61.0.42 has been allocated to ignite-b5d26a20c03fc678, duplicate allocation is not allowed 
```

innobead:
Fix #660

### DIFF
--- a/pkg/operations/remove.go
+++ b/pkg/operations/remove.go
@@ -32,15 +32,10 @@ func CleanupVM(vm *api.VM) error {
 	// Inspect the container before trying to stop it and it gets auto-removed
 	inspectResult, _ := providers.Runtime.InspectContainer(util.NewPrefixer().Prefix(vm.GetUID()))
 
-	// If the VM is running, try to kill it first so we don't leave dangling containers
-	if vm.Running() {
-		if err := StopVM(vm, true, true); err != nil {
+	// If the VM is running, try to kill it first so we don't leave dangling containers. Otherwise, try to cleanup VM networking.
+	if err := StopVM(vm, true, true); err != nil {
+		if vm.Running() {
 			return err
-		}
-	} else {
-		// Try to cleanup VM networking
-		if err := removeNetworking(util.NewPrefixer().Prefix(vm.GetUID()), false); err != nil {
-			log.Warnf("Failed to cleanup networking for stopped container %s %q: %v", vm.GetKind(), vm.GetUID(), err)
 		}
 	}
 
@@ -77,43 +72,51 @@ func RemoveVMContainer(result *runtime.ContainerInspectResult) {
 	_ = providers.Runtime.RemoveContainer(result.ID)
 }
 
-// StopVM stops or kills a VM
+// StopVM removes networking of the given VM and stops or kills it
 func StopVM(vm *api.VM, kill, silent bool) error {
 	var err error
 	container := util.NewPrefixer().Prefix(vm.GetUID())
 	action := "stop"
 
+	if !vm.Running() && !logs.Quiet {
+		log.Warnf("VM %q is not running but trying to cleanup networking for stopped container\n", vm.GetUID())
+	}
+
 	// Remove VM networking
-	if err = removeNetworking(util.NewPrefixer().Prefix(vm.GetUID()), true); err != nil {
+	if err = removeNetworking(util.NewPrefixer().Prefix(vm.GetUID())); err != nil {
+		log.Warnf("Failed to cleanup networking for stopped container %s %q: %v", vm.GetKind(), vm.GetUID(), err)
+
 		return err
 	}
 
-	// Stop or kill the VM container
-	if kill {
-		action = "kill"
-		err = providers.Runtime.KillContainer(container, signalSIGQUIT) // TODO: common constant for SIGQUIT
-	} else {
-		err = providers.Runtime.StopContainer(container, nil)
-	}
+	if vm.Running() {
+		// Stop or kill the VM container
+		if kill {
+			action = "kill"
+			err = providers.Runtime.KillContainer(container, signalSIGQUIT) // TODO: common constant for SIGQUIT
+		} else {
+			err = providers.Runtime.StopContainer(container, nil)
+		}
 
-	if err != nil {
-		return fmt.Errorf("failed to %s container for %s %q: %v", action, vm.GetKind(), vm.GetUID(), err)
-	}
+		if err != nil {
+			return fmt.Errorf("failed to %s container for %s %q: %v", action, vm.GetKind(), vm.GetUID(), err)
+		}
 
-	if silent {
-		return nil
-	}
+		if silent {
+			return nil
+		}
 
-	if logs.Quiet {
-		fmt.Println(vm.GetUID())
-	} else {
-		log.Infof("Stopped %s with name %q and ID %q", vm.GetKind(), vm.GetName(), vm.GetUID())
+		if logs.Quiet {
+			fmt.Println(vm.GetUID())
+		} else {
+			log.Infof("Stopped %s with name %q and ID %q", vm.GetKind(), vm.GetName(), vm.GetUID())
+		}
 	}
 
 	return nil
 }
 
-func removeNetworking(containerID string, isRunning bool) error {
+func removeNetworking(containerID string) error {
 	log.Infof("Removing the container with ID %q from the %q network", containerID, providers.NetworkPlugin.Name())
-	return providers.NetworkPlugin.RemoveContainerNetwork(containerID, isRunning)
+	return providers.NetworkPlugin.RemoveContainerNetwork(containerID)
 }
--- a/cmd/ignite/run/stop.go
+++ b/cmd/ignite/run/stop.go
@@ -1,8 +1,6 @@
 package run
 
 import (
-	"fmt"
-
 	api "github.com/weaveworks/ignite/pkg/apis/ignite"
 	"github.com/weaveworks/ignite/pkg/operations"
 )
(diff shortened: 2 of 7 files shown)


pkg/operations/remove.go
```diff
@@ -78,26 +75,34 @@ func RemoveVMContainer(result *runtime.ContainerInspectResult) {
...
+func StopVM(vm *api.VM, kill, silent, running bool) error {
 	var err error
 	container := util.NewPrefixer().Prefix(vm.GetUID())
 	action := "stop"
 
 	// Remove VM networking
-	if err = removeNetworking(util.NewPrefixer().Prefix(vm.GetUID()), true); err != nil {
-		return err
+	if running {
+		fmt.Printf("VM %q is not running but trying to cleanup networking for stopped container\n", vm.GetUID())
```
twelho, line 85: Could you replace this with `log.Warnf`? I feel like this should be a warning.

pkg/operations/remove.go
```diff
@@ -32,16 +32,13 @@ func CleanupVM(vm *api.VM) error {
...
+		if vm.Running() {
 			return err
 		}
-	} else {
-		// Try to cleanup VM networking
-		if err := removeNetworking(util.NewPrefixer().Prefix(vm.GetUID()), false); err != nil {
-			log.Warnf("Failed to cleanup networking for stopped container %s %q: %v", vm.GetKind(), vm.GetUID(), err)
-		}
+
+		log.Warnf("Failed to cleanup networking for stopped container %s %q: %v", vm.GetKind(), vm.GetUID(), err)
```
darkowlzz, line 41: This warning can be misleading if `StopVM` succeeds in removing network but fails to kill/stop the container.

pkg/operations/remove.go
```diff
@@ -78,26 +75,34 @@ func RemoveVMContainer(result *runtime.ContainerInspectResult) {
...
-func StopVM(vm *api.VM, kill, silent bool) error {
+func StopVM(vm *api.VM, kill, silent, running bool) error {
 	var err error
 	container := util.NewPrefixer().Prefix(vm.GetUID())
 	action := "stop"
 
 	// Remove VM networking
-	if err = removeNetworking(util.NewPrefixer().Prefix(vm.GetUID()), true); err != nil {
-		return err
+	if running {
```
darkowlzz, line 84: Shouldn't it be `!running` based on the message below ?

twelho, line 84: `!running && !logs.Quiet` to avoid printing if quiet operation is requested.

pkg/operations/remove.go
```diff
@@ -78,26 +75,34 @@ func RemoveVMContainer(result *runtime.ContainerInspectResult) {
...
+			err = providers.Runtime.KillContainer(container, signalSIGQUIT) // TODO: common constant for SIGQUIT
+		} else {
+			err = providers.Runtime.StopContainer(container, nil)
+		}
+
+		if err != nil {
+			return fmt.Errorf("failed to %s container for %s %q: %v", action, vm.GetKind(), vm.GetUID(), err)
+		}
+	} else {
+		return nil
```
darkowlzz, line 105: We can simplify this whole conditional block by returning nil if not running, reducing the level of indentation and keeping the code as it was before.

pkg/operations/remove.go
```diff
@@ -78,26 +75,34 @@ func RemoveVMContainer(result *runtime.ContainerInspectResult) {
 }
 
 // StopVM stops or kills a VM
```
twelho, line 77: Updated comment: `StopVM removes networking of the given VM and stops or kills it`

darkowlzz:
Thanks for working on this bug. This could solve the VM restart issue due to the same duplicate IP allocation issue as seen in https://github.com/weaveworks/ignite/issues/504#issuecomment-630502065 .

pkg/operations/remove.go
```diff
@@ -78,26 +75,34 @@ func RemoveVMContainer(result *runtime.ContainerInspectResult) {
...
-func StopVM(vm *api.VM, kill, silent bool) error {
+func StopVM(vm *api.VM, kill, silent, running bool) error {
 	var err error
 	container := util.NewPrefixer().Prefix(vm.GetUID())
 	action := "stop"
 
 	// Remove VM networking
```
twelho, line 83: Move this comment to the `removeNetworking` call

innobead:
@darkowlzz @twelho thanks for the review :+1:  All comments/suggestions resolved.

twelho:
Hi @innobead, thanks for your contributions! I have one more improvement to get rid of the `isRunning` boolean for removal entirely, since our usage of it was not correct in the first place. Could you apply the following patch using `git apply` and do one more commit? Thanks!

```diff
diff --git a/pkg/network/cni/cni.go b/pkg/network/cni/cni.go
index 7f953da1..c3e6877d 100644
--- a/pkg/network/cni/cni.go
+++ b/pkg/network/cni/cni.go
@@ -189,7 +189,7 @@ func cniToIgniteResult(r *gocni.CNIResult) *network.Result {
        return result
 }
 
-func (plugin *cniNetworkPlugin) RemoveContainerNetwork(containerID string, isRunning bool) (err error) {
+func (plugin *cniNetworkPlugin) RemoveContainerNetwork(containerID string) (err error) {
        if err = plugin.initialize(); err != nil {
                return err
        }
@@ -208,13 +208,10 @@ func (plugin *cniNetworkPlugin) RemoveContainerNetwork(containerID string, isRun
                return nil
        }
 
-       netnsPath := ""
-       if isRunning {
-               netnsPath = fmt.Sprintf(netNSPathFmt, c.PID)
-               if c.PID == 0 {
-                       log.Info("CNI failed to retrieve network namespace path, PID was 0")
-                       return nil
-               }
+       netnsPath := fmt.Sprintf(netNSPathFmt, c.PID)
+       if c.PID == 0 {
+               netnsPath = "" // Using a blank netns triggers a best effort cleanup
+               log.Warn("CNI failed to retrieve network namespace path, PID was 0")
        }
 
        return plugin.cni.Remove(context.Background(), containerID, netnsPath)
diff --git a/pkg/network/docker/docker.go b/pkg/network/docker/docker.go
index e3377baa..4d419267 100644
--- a/pkg/network/docker/docker.go
+++ b/pkg/network/docker/docker.go
@@ -44,7 +44,7 @@ func (plugin *dockerNetworkPlugin) SetupContainerNetwork(containerID string, _ .
        }, nil
 }
 
-func (*dockerNetworkPlugin) RemoveContainerNetwork(_ string, _ bool) error {
+func (*dockerNetworkPlugin) RemoveContainerNetwork(_ string) error {
        // no-op for docker, this is handled automatically
        return nil
 }
diff --git a/pkg/network/types.go b/pkg/network/types.go
index 89273292..f91ce053 100644
--- a/pkg/network/types.go
+++ b/pkg/network/types.go
@@ -21,7 +21,7 @@ type Plugin interface {
        SetupContainerNetwork(containerID string, portmappings ...meta.PortMapping) (*Result, error)
 
        // RemoveContainerNetwork is the method called before a container using the network plugin can be deleted
-       RemoveContainerNetwork(containerID string, isRunning bool) error
+       RemoveContainerNetwork(containerID string) error
 }
 
 type Result struct {
diff --git a/pkg/operations/remove.go b/pkg/operations/remove.go
index 1fbf907b..2786c254 100644
--- a/pkg/operations/remove.go
+++ b/pkg/operations/remove.go
@@ -83,7 +83,7 @@ func StopVM(vm *api.VM, kill, silent bool) error {
        }
 
        // Remove VM networking
-       if err = removeNetworking(util.NewPrefixer().Prefix(vm.GetUID()), true); err != nil {
+       if err = removeNetworking(util.NewPrefixer().Prefix(vm.GetUID())); err != nil {
                log.Warnf("Failed to cleanup networking for stopped container %s %q: %v", vm.GetKind(), vm.GetUID(), err)
 
                return err
@@ -116,7 +116,7 @@ func StopVM(vm *api.VM, kill, silent bool) error {
        return nil
 }
 
-func removeNetworking(containerID string, isRunning bool) error {
+func removeNetworking(containerID string) error {
        log.Infof("Removing the container with ID %q from the %q network", containerID, providers.NetworkPlugin.Name())
-       return providers.NetworkPlugin.RemoveContainerNetwork(containerID, isRunning)
+       return providers.NetworkPlugin.RemoveContainerNetwork(containerID)
 }
```

twelho:
Upon stopping a VM the container running `ignite-spawn` will be deleted, and before that happens the network is removed. We do an `InspectContainer` call in `RemoveContainerNetwork` in `pkg/network/cni/cni.go`, so the container needs to exist at that point, thus we can fill in `netnsPath`. If the PID is zero for some reason, we can queue a best effort cleanup based on the code block you linked, which should not fail even though the netns is already removed.

Since you already pass `true` to `isRunning` currently, the patch is functionally equivalent except for always calling the removal, with a blank `netnsPath` if the returned PID is 0. If the container has been externally removed, `RemoveContainerNetwork` will stop at `InspectContainer`, both before and after the patch.

innobead:
Thanks for the testing :+1:, the patch applied.

stealthybox:
LGTM 👍

PR #663, opened 2020-08-18
Add kubefire in awesome doc

innobead:


stealthybox:
This is really cool!
We got to try it out in the ignite dev call. (Feel free to join! I know it might be a bit late for your time-zone though)

Ran into some pre-requisite setup issues that we couldn't figure out on my WSL2 environment, but @darkowlzz got a 1-node kubeadm cluster running with `kubefire` on a clean digital ocean node.

We thought it was neat the cluster had Cillium by default.

innobead:
Thanks for the invitation! (ya, it should be quite late for me but I will see if I can join sometimes :smiley:  )

P.S. ignite is quite good and has a creative way to integrate with firecracker. Thanks for this project!

PR #667, opened 2020-08-23
docs/installation: set VM status.running to true

darkowlzz:
When the VM doesn't start, it confuses the users. Set it to true to
run the VM automatically.


stealthybox:
Nice -- thanks
makes `ignited` less confusing

Issue #670, opened 2020-08-23
Installation instructions dont work for Amazon Linux 2 for docker ce


Khalian:
I created an EC2 instance with an Amazon Linux 2 AMI on a r5.metal instance since I dont have a box lying around with kvm and VT-x virt lying around. 

I was looking at the instructions for installing the pre reqs here. 

https://github.com/weaveworks/ignite/blob/master/docs/installation.md

Now there are no instructions for installing ignite on Amazon Linux (or RHEL for that matter) so I tried the centos one, since that is the closest.

On top of messing up the instructions with sudo, I ran

```
which containerd || ( yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo && yum install -y containerd.io )
```

I got the error on the last command

```
[ec2-user@ip-172-31-10-187 ~]$ sudo yum install -y containerd.io
Loaded plugins: extras_suggestions, langpacks, priorities, update-motd
docker-ce-stable                                                                                      | 3.5 kB  00:00:00
(1/2): docker-ce-stable/x86_64/updateinfo                                                             |   55 B  00:00:00
(2/2): docker-ce-stable/x86_64/primary_db                                                             |  45 kB  00:00:00
Resolving Dependencies
--> Running transaction check
---> Package containerd.io.x86_64 0:1.2.13-3.2.el7 will be installed
--> Processing Dependency: container-selinux >= 2:2.74 for package: containerd.io-1.2.13-3.2.el7.x86_64
--> Finished Dependency Resolution
Error: Package: containerd.io-1.2.13-3.2.el7.x86_64 (docker-ce-stable)
           Requires: container-selinux >= 2:2.74
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```

Now I could go and start hunting for how to install containerd in AL2, but I expect a company like weaveworks to have a more elaborate set of instructions for installation, including reliable reproducible ways to install even the pre reqs.

Khalian:
This should cover it : https://github.com/weaveworks/ignite/pull/671

stealthybox:
Sorry -- I missed this issue when reading your PR.
It looks like this require is failing:
```
Requires: container-selinux >= 2:2.74
```

Here's the fix, I'll update the PR suggestion
https://github.com/aws/containers-roadmap/issues/614#issuecomment-559619366

PR #671, opened 2020-08-24
Update installation instructions for Amazon Linux

Khalian:
Not everyone has a machine/instance lying around with kvm and virt assist. A lot of developers just leverage cloud providers to give them the necessary instances to work it.

This commit improves installation instructions for Amazon Linux 2

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -42,6 +42,8 @@ KVM acceleration can be used
 
 With this kind of output, you're ready to go!
 
+Alternatively if you want to run this on AWS EC2, use Amazon Linux v2 (https://aws.amazon.com/amazon-linux-2/) and ensure you use a baremetal instance, like the r series like say r5.metal (https://aws.amazon.com/ec2/instance-types/). 
+
 ## Installing dependencies
 
 Ignite has a few dependencies (read more in this [doc](dependencies.md)).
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -64,6 +66,14 @@ which containerd || ( yum-config-manager --add-repo https://download.docker.com/
     # Install containerd if it's not present
 ```
 
+Amazon Linux 2
+
+```bash
+yum install -y e2fsprogs openssh-clients git 
+which containerd || sudo yum install -y docker
```
stealthybox, line 73: Thanks for the patch -- this line should ideally just install `containerd`, not a full install of `docker`.

There doesn't appear to be a docker or containerd package on a default AmazonLinux 2 system.
*( I only checked with the `amazonlinux:2` docker container's default yum repo setup though )

If you add the docker-ce repo directly from docker, it's possible to install the most up-to-date version of containerd from there, just like the CentOS example:
```suggestion
which containerd || ( yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo && yum install -y containerd.io )
```

Khalian: That was precisely the command I ran before and it did not work. Your suggested changes dont work. Refer to https://github.com/weaveworks/ignite/issues/670.

stealthybox: Ah thanks! Sorry about that -- here's a better fix:
```suggestion
amazon-linux-extras enable docker
yum install containerd
```

I was able to verify this works within an AL2 docker container

stealthybox:
Fixes #670

docs/installation.md
Khalian: So yes that works. And that would work fine with docker hub stuff. But when I try it with ECR I cant get it to work : https://github.com/weaveworks/ignite/issues/673

Khalian: Anyways I still applied it.

stealthybox: Thanks for making the install uniform for containerd.
Also appreciate the other bug you opened about registry auth.

I'm glad the docker runtime is helping you get around that :+1:

Khalian: Yea so about that, I think we need some instructions for runtime modes and more commands like ignite pull. I will be happy to write that down, but where should I put it?

PR #676, opened 2020-08-27
Update ignite help, image and kernel to no longer require root

Khalian:
Help should not need root.

### DIFF
--- a/cmd/ignite/cmd/root.go
+++ b/cmd/ignite/cmd/root.go
@@ -43,7 +43,7 @@ func NewIgniteCommand(in io.Reader, out, err io.Writer) *cobra.Command {
 
 			// TODO Some commands do not need to check root
 			// Currently it seems to be only ignite version that does not require root
-			if cmd.Name() == "version" && cmd.Parent().Name() == "ignite" {
+			if isNonRootCommand(cmd.Name(), cmd.Parent().Name()) {
 				return
 			}
 
@@ -115,6 +115,19 @@ func NewIgniteCommand(in io.Reader, out, err io.Writer) *cobra.Command {
 	root.AddCommand(NewCmdStop(os.Stdout))
 	root.AddCommand(versioncmd.NewCmdVersion(os.Stdout))
 	return root
+}
+
+func isNonRootCommand(cmd string, parentCmd string) bool {
+	if parentCmd != "ignite" {
+		return false
+	}
+
+	switch cmd {
+	case "version", "help", "image", "kernel", "completion", "inspect", "ps":
+		return true
+	}
+
+	return false
 }
 
 func addGlobalFlags(fs *pflag.FlagSet) {


Khalian:
Testing

```
make
make install
```

Then ran

```
ignite help
ignite image ps
ignite kernel ps
```

And all of them passed without a hitch.

cmd/ignite/cmd/root.go
```diff
@@ -117,6 +117,14 @@ func NewIgniteCommand(in io.Reader, out, err io.Writer) *cobra.Command {
 	return root
 }
 
+func isNonRootCommand(cmd string, parentCmd string) bool {
+	isVersionCmd := cmd == "version" && parentCmd == "ignite"
+	isHelpCmd := cmd == "help" && parentCmd == "ignite"
+	isImageCmd := cmd == "image" && parentCmd == "ignite"
+	isKernelCmd := cmd == "kernel" && parentCmd == "ignite"
+	return isVersionCmd || isHelpCmd || isImageCmd || isKernelCmd
```
darkowlzz, line 125: This could be simplified using a switch statement. Something like:

```go
if parentCmd != "ignite" {
        return false
}

switch cmd {
    case "version", "help", "image", "kernel":
        return true
}

return false
```

darkowlzz:
Thanks for working on this. I went through all the commands to see which of them don't require root.
`ps`, `inspect` and `completion` don't require root.
`inspect` for `vm`, `image` and `kernel`, all work without root as they are just reading the metadata file.
Most of the other commands require permission to access some files and interact with the runtime.

stealthybox:
This patch already looks good but +1 on using the switch and adding the other commands if you have a moment.

Thanks for fixing this!

Khalian:
> This patch already looks good but +1 on using the switch and adding the other commands if you have a moment.
> 
> Thanks for fixing this!

I dont understand, what more commands you want me to patch here?

cmd/ignite/cmd/root.go
Khalian: Done

cmd/ignite/cmd/root.go
```diff
@@ -117,6 +117,19 @@ func NewIgniteCommand(in io.Reader, out, err io.Writer) *cobra.Command {
 	return root
 }
 
+func isNonRootCommand(cmd string, parentCmd string) bool {
+	if parentCmd != "ignite" {
+		return false
+	}
+
+	switch cmd {
+	case "version", "help", "image", "kernel", "completion", "inspect":
```
darkowlzz, line 126: Can you also ass `ps`?

darkowlzz (commented):
Manually tested it and works well.
Almost ready, just need to include `ps` as well.

Khalian:
> Manually tested it and works well.
> Almost ready, just need to include `ps` as well.

Done. I made the change.

darkowlzz (approved):
LGTM! Thanks a lot for the patch.

PR #678, opened 2020-08-31
Fixes command in docs

gabeio:
changes command slightly to fix markdown issue.

Markdown was misunderstanding the ` to be a markdown instead of command. I changed it slightly to do the same thing yet not break any markdown.


darkowlzz (approved):
Thanks for fixing that. LGTM.
The CircleCI failure is due to availability of dependency updates and the TravisCI failure is due to a flaky test that we have observed before. Created #679 to track the issue.

PR #686, opened 2020-09-06
Export command option types

darkowlzz:
Ignite command functions accept unexported types like `startOptions`,
`createOptions`, etc. This makes the cmd/ignite/run package unusable for
projects that import ignite.
IDEs warn about unexported return type for functions like
`NewStartOptions()`, `NewCreateOptions()`, etc.
This change makes those option types public.


stealthybox:
I'm fine with this, especially since our implementations are currently too conflated with the cobra command setup code.

We don't promise these types will be stable, but folks can pin to our releases, and exposing them allows others to build cool things.

PR #669, opened 2020-08-23
Make runtime and network-plugin VM create flags

darkowlzz:
This is a follow-up change to https://github.com/weaveworks/ignite/pull/651, which adds the runtime and
network-plugin info in the VM status.

- Remove the global flags `runtime` and `network-plugin` to be in VM `create`
command.
- Update all the VM commands (`run, stop, rm`) to read the `runtime` and
`network-plugin` info from VM status.
- Add `runtime` flag to `image import`.
- Update docs.
- Handle rm for stopped VMs - stopped VMs don't contain runtime info. Avoid runtime operations for them.
- Fix `patchStopped()` status patch data in ignite-spawn as per the recent status changes.
- Update e2e tests.
- Handle VM start after stop - since stopped VMs don't contain runtime info, add runtime and network flags
to VM start command.
- Add e2e tests for VM lifecycle, including VM provider switching.

PR #687, opened 2020-09-12
Fix starting VM with non-default providers

darkowlzz:
At VM start, check if the providers flags are set before updating the
VM status with the providers info. This ensures that the providers of a
VM are changed only when explicitly set in the start command.

Add e2e test to verify the fix.

Follow-up to #669


stealthybox:
ty for bug fix friend :+1:

PR #697, opened 2020-09-26
Add Publish OS Image github action

darkowlzz:
Publish OS Image action is a manually triggered action with parameters
GOARCH, WHAT, RELEASE, USER for building and publishing OS container
image.
It is a developer automation action for building and publishing OS
images easily.

**NOTE**: Must set `CR_PAT` and `CR_HOST` in the github repo secrets for
the image push to succeed.

Tested the action in a fork for building various images and publishing to
github container registry and manually verified that the published images
boot and are usable.
Refer:
- https://github.com/darkowlzz?tab=packages
- https://github.com/darkowlzz/ignite/actions/runs/273283567

Image build for alpine fails due to missing alpine.tar file. This can be
fixed in the makefile separately.
~~Opensuse image build fails because the base image for it isn't public and
can't be pulled, [logs](https://github.com/darkowlzz/ignite/runs/1168464264?check_suite_focus=true#step:3:25).~~ Works with correct RELEASE value, was using `latest`.

### DIFF
--- a/.github/workflows/publish-os-image.yml
+++ b/.github/workflows/publish-os-image.yml
@@ -1,0 +1,54 @@
+name: Publish OS Image
+
+on:
+  workflow_dispatch:
+    # Enable manual trigger of this action.
+    inputs:
+      arch:
+        description: Platform architecture (amd64, arm64).
+        default: amd64
+        required: true
+      what:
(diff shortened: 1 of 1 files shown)


darkowlzz:
Closing for #717

PR #704, opened 2020-09-28
Update CODEOWNERS -- fix bug + add @darkowlzz

stealthybox:
Sunny noticed these need to be on the same line.
Each subsequent line was overriding the previous ones which meant @twelho was assigned for everything! 🙃


darkowlzz (approved):
LGTM!

PR #712, opened 2020-10-12
Fix concurrent VM creation due to busy device error

darkowlzz:
Move the lock ignite snapshot lock from `ActivateSnapshot()` to
`copyToOverlay()` in the same `pkg/dmlegacy` package. This guards against
concurrent interactions with the device mapper by serializing it with
the snapshot lock file.

TODO: Add an e2e test.


chanwit (approved):
LGTM
Thank you @darkowlzz

stealthybox:
lgtm pending tests

stealthybox:
We were able to create some hard-to-reproduce / transient lock failures on @darkowlzz's DigitalOcean dev machine.

I was not able to create any lockfile issues after testing ~100 vm's in small and medium sized concurrent batches on my WSL2 laptop.

This seems to be working about as stable as it did before :+1:
e2e test coming next week using `cmd.Start` and `cmd.Wait` in two loops.

PR #715, opened 2020-10-18
e2e: Add test for concurrent VM creation

darkowlzz:
Run multiple VMs together and wait for the commands to finish successfully. Cleanup at the end.
Test file name has a `z` prefix to ensure it never runs at first. If the images aren't already available on the host, all the VM creation processes will try to pull the images at the same time and corrupt the image store.

Follow up of #712 .


stealthybox:
Nice test!
very clean and easy to read :)

Issue #329, opened 2019-08-12
Add a k3s sample image


luxas:
cc @alexellis :)

alexellis:
This sounds awesome 😄

chanwit:
Current version of k3s is now v0.8.1: https://github.com/rancher/k3s/releases/tag/v0.8.1

EmeraldShift:
Hi! @doracgp and I are students at the University of Texas at Austin, and we are interested in virtualization and containerization. We'd like to tackle this issue. Can it be assigned to us?

To be sure, what does this issue entail, beyond adding a functional k3s image to `images/`?

stealthybox:
@EmeraldShift just the new image + some edits to the Makefile should be perfect.
If you'd like to update or author some related documentation, that could be a beneficial second step.

Feel free to take this one!  It'd be a great addition :+1:

stealthybox:
Feel free to join our dev call on Mondays if you need help :)

EmeraldShift:
This issue can be closed, as #739 resolves this.

PR #739, opened 2020-11-26
Add a k3s sample image

EmeraldShift:
This change adds an image for k3s, a lightweight Kubernetes distribution, as per #329.

A few notes:
* The `rancher/k3s` image seems to be too minimal to function correctly—for example, it wouldn't mount `/proc`, etc. Thus, this image is based on Ubuntu 20.04, with k3s installed via the provided Rancher script
* Docker can't properly handle [tags with a `+` symbol in them](https://github.com/moby/moby/issues/16304), so the version of k3s used for this image is the highest version before `+` was introduced into the release tags, i.e. v1.0.1.
* The image was tested by deploying a [hello-kubernetes](https://github.com/paulbouwer/hello-kubernetes) manifest. The service was reachable from the host (outside the VM) via the external IP leased to the `hello-kubernetes` Service.

### DIFF
--- a/.github/workflows/update-go-deps.yml
+++ b/.github/workflows/update-go-deps.yml
@@ -25,7 +25,7 @@ jobs:
           make go-autogen
       - name: Create pull request
         id: cpr
-        uses: peter-evans/create-pull-request@v3.5.0
+        uses: peter-evans/create-pull-request@v3.5.1
         with:
           commit-message: "[automated] Update go dependencies"
           branch: go-deps-patch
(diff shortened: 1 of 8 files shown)


EmeraldShift:
@stealthybox I think this one's ready for review.
Looking through the details of the build failure, I'm not sure how they're related to these changes, but instead they seem like transient failures of some sort. I'm not sure how to get them to pass. Any ideas?

darkowlzz (commented):
I tried the image myself and it works as expected.
Had to set 2GB memory for the VM, else k3s fails with out of memory error. The default 512MB of memory is not enough. We can document it better, similar to the README in images/kubeadm/README.md.
We can also add docs for multi-node HA setup separately, similar to kubeadm, in the future.
Thanks for working on this.

darkowlzz:
We don't run the OS image builds in the regular CI, so the build failure is unrelated. Don't need to worry about it 🙂

EmeraldShift:
@darkowlzz Thank you for the feedback! I've patched the Makefile with the changes you mentioned, and added a simple README.

stealthybox:
LGTM -- thanks for fixing the tag problem

darkowlzz (approved):
LGTM! Thanks.

PR #741, opened 2020-12-01
Fix MASTER_IP definition in kubeadm README

### LINKED ISSUE #740: Malformed syntax & outdated/invalid `jq` argument in kubeadm image README
`images/kubeadm/README.md` fetches the IP address of the `master-0` VM with the following line:
```bash
export MASTER_IP=$($ignite inspect vm master-0 | jq -r ".status.ipAddresses[0]")
```

This seems to be problematic, since there's an extra `$` before `ignite`, which (unless set) evaluates to an empty string, causing the shell to run `inspect ...`, to no avail.

Additionally, the JSON path `.status.ipAddresses[0]` seems either invalid or outdated, as the correct path appears to be `.status.network.ipAddresses[0]` as of the time of writing.

EmeraldShift:
Resolves #740. Tested by spawning a VM with name `master-0`, running the command, and reading `$MASTER_IP`.

```bash
ignite run ... --name master-0
export MASTER_IP=$(ignite inspect vm master-0 | jq -r ".status.network.ipAddresses[0]")
echo $MASTER_IP # Prints a valid IP address.
```

### DIFF
--- a/images/kubeadm/README.md
+++ b/images/kubeadm/README.md
@@ -29,7 +29,7 @@ ignite run weaveworks/ignite-kubeadm:latest \
     --name master-0
 
 # Get the IP address of the initial master, for the kubeadm join command below
-export MASTER_IP=$($ignite inspect vm master-0 | jq -r ".status.ipAddresses[0]")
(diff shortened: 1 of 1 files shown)


stealthybox:
Just tried this edit out a little bit -- thanks for the fix :)

PR #759, opened 2020-12-23
Show output in test to aid debugging

bboreham:
Otherwise when it fails we just get something like `exit status 1`


darkowlzz (approved):
LGTM!
Very helpful. Thanks.

PR #760, opened 2020-12-23
Small test improvements

bboreham:
Mark `Run()` as helper function, so failure will be reported on the line of the caller.

Also fix a couple of typos in the same file.


darkowlzz (approved):
LGTM! Thanks.

PR #762, opened 2020-12-23
Refactor: extract TarExtract function

bboreham:
Same block of 15-20 lines was repeated three times.
This version is more consistent, calling Reader.Close() every time.


darkowlzz (approved):
LGTM!

bboreham:
Note I'm not merging this because Travis CI hasn't run.

PR #763, opened 2020-12-23
Limit Travis branch builds to just master

bboreham:
PRs always get sent to Travis by GitHub integration, so we don't need Travis to do the identical build for the branch too.

Once the PR is merged, that will be a commit on master, which we do want Travis to build.

Found at https://github.com/travis-ci/travis-ci/issues/1147


bboreham:
This blocked all travis CI, which is not what I wanted.

Issue #679, opened 2020-08-31
flaky volume e2e test when using docker + CNI


darkowlzz:
The volume e2e test is observed to fail frequently with Docker + CNI :

```console
=== RUN   TestVolumeWithDockerAndCNI

    TestVolumeWithDockerAndCNI: run_volume_test.go:120: assertion failed: error is not nil: exit status 1: vm stop: 

        ["/home/travis/build/weaveworks/ignite/bin/ignite" "--runtime=docker" "--network-plugin=cni" "stop" "e2e_test_volume_docker_and_cni"]

        time="2020-08-31T04:45:29Z" level=info msg="Removing the container with ID \"ignite-d54e77eeaa408deb\" from the \"cni\" network"

        time="2020-08-31T04:45:31Z" level=fatal msg="failed to stop container for VM \"d54e77eeaa408deb\": Error response from daemon: No such container: ignite-d54e77eeaa408deb"

```

It passes on restarting the test, mostly.

Docker + CNI never worked on my development machine (ubuntu 18.04) but it works on TravisCI for all of the other e2e tests.

bboreham:
Similar failure in `TestVMLifecycleWithDockerAndDockerBridge` https://travis-ci.com/github/weaveworks/ignite/builds/210384312

bboreham:
I drilled into this a bit on SemaphoreCI - can't get the test to fail reliably on my own machine.
It appears to be a race inside `ContainerStop()` whereby the container is removed before the operation completes.
Container removal is expected since Ignite sets `AutoRemove` (unless debug is on).

Evidence: I ran `docker events` in background and manually executed the `ignite` commands from the test:

```
semaphore@semaphore-vm:~/ignite$ sudo bin/ignite stop e2e-test-vm-lifecycle-docker-and-cni
INFO[0000] Removing the container with ID "ignite-fb4ced5044c57fb1" from the "cni" network 
2021-01-14T18:41:42.873403421Z container kill 691a608a71393439c6c38663730f5d86e9d1c8a538ceaff80a8b898319f9ee21 (ignite.name=e2e-test-vm-lifecycle-docker-and-cni, image=docker.io/weaveworks/ignite:dev, name=ignite-fb4ced5044c57fb1, signal=15)
[...]

Issue #658, opened 2020-08-10
TODO Fix `ignite stop --runtime=docker` related CI test failure


stealthybox:
See the commented out `TestVolumeWithDockerAndDockerBridge` from #653 for the unrelated test failure.

@darkowlzz mentions that this test passes on his digital ocean machine.
One guess is that this failure is timing or race related.
We suppose that this could be resolved in the docker runtime client.

bboreham:
Isn't this the same problem as in #679 ?

darkowlzz:
> Isn't this the same problem as in #679 ?

Yes, they are all docker related with the same container stop issue. The fix in #773 should solve this.
Great job on that 🥇

darkowlzz:
> did #773 solve this?

Yes, that solved the issue and since then we have been running `TestVolumeWithDockerAndDockerBridge` test without any issue.
Closing this issue. Feel free to ask any related questions if you have any. Thanks :)

PR #772, opened 2021-01-15
Run lifecycle and volume tests with --debug

bboreham:
Work around https://github.com/weaveworks/ignite/issues/679, but leaves dead containers lying around after test.

Also #658


bboreham:
Replaced by #773

PR #796, opened 2021-03-03
Use declarative port mappings and copy files values when CLI flags are empty

### LINKED ISSUE #795: Declarative port mappings and copy files values are getting cleared while CLI flags are empty
The title is self-explanatory but I should give more details.

I start an ignite VM using the below declarative configuration by executing `ignite run --config ubuntu.yaml`:

```yaml
apiVersion: ignite.weave.works/v1alpha3
kind: VM
metadata:
  name: ubuntu
spec:
  image:
    oci: weaveworks/ignite-ubuntu
  cpus: 1
  diskSize: 400MB
  memory: 64MB
  network:
    ports:
    - hostPort: 2222
      vmPort: 22
  ssh: true
``` 

When I run `ignite ps` I don't see any port mappings on the VM.

I've done some debugging and found where the issue is and it also showed me that `copyFiles` suffers the same problem.

Here's the offending line: https://github.com/weaveworks/ignite/blob/3120baf0e4f924a9280d59e7092fda3c57bcccb1/cmd/ignite/run/create.go#L221

I'll be creating a PR to fix this issue

onatm:
Fixes #795

### DIFF
--- a/cmd/ignite/run/create.go
+++ b/cmd/ignite/run/create.go
@@ -217,10 +217,12 @@ func applyVMFlagOverrides(baseVM *api.VM, cf *CreateFlags, fs *flag.FlagSet) err
 		return err
 	}
 
-	// Parse the given port mappings.
-	baseVM.Spec.Network.Ports, err = meta.ParsePortMappings(cf.PortMappings)
-	if err != nil {
-		return err
+	if len(cf.PortMappings) > 0 {
+		// Parse the given port mappings.
+		baseVM.Spec.Network.Ports, err = meta.ParsePortMappings(cf.PortMappings)
+		if err != nil {
+			return err
+		}
 	}
 
 	// If the SSH flag was set, copy it over to the API type
--- a/cmd/ignite/run/create.go
+++ b/cmd/ignite/run/create.go
@@ -211,10 +211,12 @@ func applyVMFlagOverrides(baseVM *api.VM, cf *CreateFlags, fs *flag.FlagSet) err
 		baseVM.Spec.Storage = cf.VM.Spec.Storage
 	}
 
-	// Parse the --copy-files flag.
-	baseVM.Spec.CopyFiles, err = parseFileMappings(cf.CopyFiles)
-	if err != nil {
-		return err
+	if len(cf.CopyFiles) > 0 {
+		// Parse the --copy-files flag.
+		baseVM.Spec.CopyFiles, err = parseFileMappings(cf.CopyFiles)
+		if err != nil {
+			return err
+		}
 	}
 
 	if len(cf.PortMappings) > 0 {
--- a/cmd/ignite/run/create_test.go
+++ b/cmd/ignite/run/create_test.go
@@ -220,6 +220,27 @@ func TestApplyVMFlagOverrides(t *testing.T) {
 			err: true,
 		},
 		{
+			name: "copy files in baseVM",
+			createFlag: &CreateFlags{
+				VM: &api.VM{
+					Spec: api.VMSpec{
+						CopyFiles: []api.FileMapping{
+							{
+								HostPath: "/tmp/foo",
+								VMPath:   "/tmp/bar",
+							},
+						},
+					},
+				},
+			},
+			wantCopyFiles: []api.FileMapping{
+				{
+					HostPath: "/tmp/foo",
+					VMPath:   "/tmp/bar",
+				},
+			},
+		},
+		{
 			name: "valid port mapping",
 			createFlag: &CreateFlags{
 				VM:           &api.VM{},
@@ -241,6 +262,33 @@ func TestApplyVMFlagOverrides(t *testing.T) {
 				PortMappings: []string{"1.1.1.1:foo:bar"},
 			},
 			err: true,
+		},
+		{
+			name: "port mapping in base VM",
+			createFlag: &CreateFlags{
+				VM: &api.VM{
+					Spec: api.VMSpec{
+						Network: api.VMNetworkSpec{
+							Ports: meta.PortMappings{
+								{
+									BindAddress: net.IPv4(0, 0, 0, 0),
+									HostPort:    uint64(80),
+									VMPort:      uint64(80),
+									Protocol:    meta.ProtocolTCP,
+								},
+							},
+						},
+					},
+				},
+			},
+			wantPortMapping: meta.PortMappings{
+				meta.PortMapping{
+					BindAddress: net.IPv4(0, 0, 0, 0),
+					HostPort:    uint64(80),
+					VMPort:      uint64(80),
+					Protocol:    meta.ProtocolTCP,
+				},
+			},
 		},
 		{
 			name: "ssh public key set",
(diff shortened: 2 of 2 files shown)


onatm:
I've just found the following assertions are completely wrong:

https://github.com/weaveworks/ignite/blob/3120baf0e4f924a9280d59e7092fda3c57bcccb1/cmd/ignite/run/create_test.go#L285-L293

onatm:
`TestApplyVMFlagOverrides` is written to validate the wrong implementation and I strongly recommend to rewrite it.

cmd/ignite/run/create.go
```diff
@@ -211,16 +211,22 @@ func applyVMFlagOverrides(baseVM *api.VM, cf *CreateFlags, fs *flag.FlagSet) err
...
+			return err
+		}
 	}
 
-	// Parse the given port mappings.
-	baseVM.Spec.Network.Ports, err = meta.ParsePortMappings(cf.PortMappings)
-	if err != nil {
-		return err
+	// TODO: Remove len(cf.PortMappings) > 0 after fixing TestApplyVMFlagOverrides
+	if fs.Changed("ports") || len(cf.PortMappings) > 0 {
```
darkowlzz, line 224: Checking the length for both CopyFiles and PortMappings should be enough to get this right.
"copy-files" and "ports" are not directly taken from the VM spec. They have separate variables for simpler flag friendly types, refer: https://github.com/weaveworks/ignite/blob/3120baf0e4f924a9280d59e7092fda3c57bcccb1/cmd/ignite/cmd/vmcmd/create.go#L71-L72 .

darkowlzz (commented):
Thanks for reporting and fixing this bug.

> TBH, I cannot see any value in TestApplyVMFlagOverrides because it is not actually testing the override logic as it happens in the CLI.

Some background for `applyVMFlagOverrides()` would help understand it better 🙂 
Before this, we used to have [`TestConstructVMFromCLI()`](https://github.com/weaveworks/ignite/blob/d1d003683db7541559505e5a39cbfbe640ad893e/cmd/ignite/run/create_test.go#L16), that did exactly what you're looking for.
`applyVMFlagOverrides()` was introduced to support multiple layers of configurations - component config (global ignite configuration), VM config (set via `--config` run command flag) and flag configuration (like `--copy-flags` and `--ports`). The base VM that this function receives is after combing the component config and VM config. This function tries to capture the user's intent expressed via flags to set the properties of the VM. The flags in ignite run command is a combination of simple types and VM spec type. The types that are complex to be in flag have a simpler form that is overwritten in the VM spec.  [`CreateFlag`](https://github.com/weaveworks/ignite/blob/master/cmd/ignite/run/create.go#L32-L42) contains `api.VM` for `VM` and separate fields for `SSH`, `PortMappings` and `CopyFiles`. `api.VM` contains ssh, portmappings and copyfiles, but aren't flag friendly so they have separate types to simplify that.
Because of this combination of `api.VM`, simple types for some properties, and merging of base configurations, it's difficult to find out the user's intent in the baseVM and the CreateFlags. One way to know if the user specified a flag is to do `fs.Changed()` on the fields that the users can set and are also part of api.VM. When a change is found, the value in baseVM is overwritten by the flag value.
In case of CopyFiles and PortMappings, since they have separate variables for their flags, we don't need to do `fs.Changed()`. If there's any value, we should just use it. So just checking their length before parsing them should fix that issue.
Hope this helps understand how we have to apply the baseVM VM configurations from two different sources with the VM config expressed via the CLI flags.

Regarding the tests, thanks for noticing the incorrect assertions and fixing them and improvements in the comparison.
I would recommend adding a few cases as follows to validate that the CopyFiles and PortMappings set in the baseVM from a config file don't get overwritten by empty flags for them:

```go
                {
                        name: "copy files in baseVM",
                        createFlag: &CreateFlags{
                                VM: &api.VM{
                                        Spec: api.VMSpec{
                                                CopyFiles: []api.FileMapping{
                                                        {
                                                                HostPath: "/tmp/foo",
                                                                VMPath:   "/tmp/bar",
                                                        },
                                                },
                                        },
                                },
                        },
                        wantCopyFiles: []api.FileMapping{
                                {
                                        HostPath: "/tmp/foo",
                                        VMPath:   "/tmp/bar",
                                },
                        },
                },
...
                {
                        name: "port mapping in base VM",
                        createFlag: &CreateFlags{
                                VM: &api.VM{
                                        Spec: api.VMSpec{
                                                Network: api.VMNetworkSpec{
                                                        Ports: meta.PortMappings{
                                                                {
                                                                        BindAddress: net.IPv4(0, 0, 0, 0),
                                                                        HostPort:    uint64(80),
                                                                        VMPort:      uint64(80),
                                                                        Protocol:    meta.ProtocolTCP,
                                                                },
                                                        },
                                                },
                                        },
                                },
                        },
                        wantPortMapping: meta.PortMappings{
                                meta.PortMapping{
                                        BindAddress: net.IPv4(0, 0, 0, 0),
                                        HostPort:    uint64(80),
                                        VMPort:      uint64(80),
                                        Protocol:    meta.ProtocolTCP,
                                },
                        },
                },
```

With these and your improvements in the test assertions, we should have the right tests that covers this bug.
Maybe we should also support appending the port mappings and copy files from the config file and CLI, but we can work on that separately.

Hope that helps understand it better.

onatm:
@darkowlzz for your detailed explanation about how `ignite` handles different sets of configurations. Now I have a better understanding of how the config values flow through the application 👍 I'll make the changes you recommended and finish working on this PR.

cmd/ignite/run/create.go
onatm: I am going to drop the commit https://github.com/weaveworks/ignite/pull/796/commits/c5d81b93f9ba34abc4c6b11afea46ddddbe9573d

cmd/ignite/run/create_test.go
```diff
@@ -273,23 +273,23 @@ func TestApplyVMFlagOverrides(t *testing.T) {
...
-						t.Errorf("expected VM.Spec.CopyFiles to be %v, actual: %v", rt.wantCopyFiles, rt.createFlag.VM.Spec.CopyFiles)
+					if len(rt.wantCopyFiles) != len(vm.Spec.CopyFiles) {
+						t.Errorf("expected VM.Spec.CopyFiles to be %v, actual: %v", rt.wantCopyFiles, vm.Spec.CopyFiles)
 					}
 				}
 
 				// Check if port mappings are set as expected.
-				if reflect.DeepEqual(rt.createFlag.VM.Spec.Network.Ports, rt.wantPortMapping) {
-					t.Errorf("expected VM.Spec.Network.Ports to be %v, actual: %v", rt.wantPortMapping, rt.createFlag.VM.Spec.Network.Ports)
+				if vm.Spec.Network.Ports.String() != rt.wantPortMapping.String() {
```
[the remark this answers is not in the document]
onatm: I'll amend the commit to improve the error message

onatm:
I've just pushed the changes you recommended @darkowlzz Thank you for the review!

darkowlzz (approved):
Thanks a lot 🙂 
LGTM!

PR #799, opened 2021-03-05
Bump ignite-spawn base image to Alpine to 3.13

gaby:
Alpine 3.13 was released on 2021-01-14
Release notes: https://alpinelinux.org/posts/Alpine-3.13.0-released.html


stealthybox:
@gaby if you're looking to bump the Alpine OS image as well, see the images directory

Thanks for this update to the sandbox container

gaby:
Will take a look at that tonight.

stealthybox:
This may be a challenge to update -- we may need to find a better rootfs source for alpine that is versioned:

https://github.com/weaveworks/ignite/blob/1f0faef/images/alpine/Makefile#L8

Issue #768, opened 2021-01-09
where to get wireguard kernel with k3s ubuntu 20.04 


matti:
error: Module wireguard not found in directory /lib/modules/5.4.43

```
$ ignite run weaveworks/ignite-ubuntu:20.04-amd64 --kernel-image weaveworks/ignite-kernel:5.4.43 --cpus 32 --ssh --memory 4GB --size 10GB --ssh
```

```
$ apt install wireguard
Reading package lists... Done
Building dependency tree
Reading state information... Done
wireguard is already the newest version (1.0.20200513-1~20.04.2).

$ curl -sfL https://get.k3s.io | K3S_URL=https://master:6443 K3S_TOKEN=token  sh -
```

```
$ journalctl -f -u k3s-agent
```
>failed to run command: export SUBNET_IP=$(echo $SUBNET | cut -d'/' -f 1); ip link del flannel.1 2>/dev/null; echo $PATH >&2; wg-add.sh flannel.1 && wg set flannel.1 listen-port 51820 private-key privatekey && ip addr add $SUBNET_IP/32 dev flannel.1 && ip link set flannel.1 up && ip route add $NETWORK dev flannel.1 Err: exit status 1 Output: /var/lib/rancher/k3s/data/986d5e8cf570f904598f9a5d531da2430e5a6171d22b7addb1e4a7c5b87a47d0/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/var/lib/rancher/k3s/data/986d5e8cf570f904598f9a5d531da2430e5a6171d22b7addb1e4a7c5b87a47d0/bin/aux\nmodprobe: FATAL: Module wireguard not found in directory /lib/modules/5.4.43\nError: Unknown device type.\n/var/lib/rancher/k3s/data/986d5e8cf570f904598f9a5d531da2430e5a6171d22b7addb1e4a7c5b87a47d0/bin/aux/wg-add.sh: line 26: boringtun: command not found\n/var/lib/rancher/k3s/data/986d5e8cf570f904598f9a5d531da2430e5a6171d22b7addb1e4a7c5b87a47d0/bin/aux/wg-add.sh: line 29: boringtun: command not found\n/var/lib/rancher/k3s/data/986d5e8cf570f904598f9a5d531da2430e5a6171d22b7addb1e4a7c5b87a47d0/bin/aux/wg-add.sh: line 32: wireguard-go: command not found"
```

matti:
[...]

Issue #623, opened 2020-06-15
Bump default kernel to 5.4 LTS


twelho:
The 5.4 LTS series is now the latest LTS release available and has had time to stabilize. It should work with all VM images available (Ubuntu 18.04 LTS has it available in the `linux-generic-hwe-18.04-edge` package, so it should work as well).

gaby:
@twelho I was wondering the same thing, it seems all the kernels/images from ignite are outdated.

PR #806, opened 2021-03-10
Bump default kernel to 5.4.102

gaby:
Now that new kernel images are being built, the default kernel used by Ignite needs to be updated.

There's also several tickets regarding bumping this to kernel 5.4.x, thoughts?

See #768 #623

### DIFF
--- a/cmd/ignite/run/testdata/output/apply-vm-config-empty.json
+++ b/cmd/ignite/run/testdata/output/apply-vm-config-empty.json
@@ -13,7 +13,7 @@
       "oci": "weaveworks/ignite:dev"
     },
     "kernel": {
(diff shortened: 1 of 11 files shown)


gaby:
It seems like the tests also need to be updated. I will take a look again tonight.

gaby:
Seems like the test needs to be re-run? @stealthybox

stealthybox:
@gaby -- after talking this through /w @darkowlzz we think your suggestion to update the default to `5.4.102` is a good one.
Would you like to update this patch? :)

gaby:
@stealthybox I just pushed the requested change. Default kernel is now 5.4.102

PR #823, opened 2021-04-18
Ignore not found error while deactivating snapshot

darkowlzz:
While deleting a VM, if the associated device mapper device has been
already deleted and cannot be found, the delete should continue without
any error.

Based on:

```console
sudo bin/ignite vm rm -f young-surf
FATA[0000] command ["dmsetup" "remove" "--verifyudev" "ignite-bb967b1753deda1e"] exited with "device-mapper: remove ioctl on ignite-bb967b1753deda1e  failed: No such device or address\nCommand failed.\n": exit status 1
```

### DIFF
--- a/pkg/dmlegacy/deactivate.go
+++ b/pkg/dmlegacy/deactivate.go
@@ -4,12 +4,16 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


stealthybox:
Verified locally as well

Issue #832, opened 2021-05-07
Re-work network connectivity


networkop:
Hi again 👋 ,
I've been looking at how network plumbing has been done and I think it can be improved slightly.
Right now networking is setup _after_ the container is started. 
```
providers.Runtime.RunContainer
...
providers.NetworkPlugin.SetupContainerNetwork
```
Followed by the `ignite-spawn` 's call to `SetupContainerNetworking` which waits for at least one interface to be connected.

The problem is that, in theory, this could lead to non-deterministic networking setup. Let's assume that we have some CNI plugin connecting more than one interface at the same time as `wait.PollImmediate` in `pkg/container/network.go` tries to setup its networks. Because these two processes happen in parallel, it's possible to either see all interfaces  or only a subset of them, if CNI hasn't connected them all yet. 

So an alternative solution could be something like this:

1. Create a container during `ignite create` -- this would require an extra function in container Interface
2. Connect all the interfaces with CNI/CNM right after this step
3. Change the behaviour of `ignite start` to just call `RunContainer` -- obv, dropping the option for user to reconfigure VMs on the start.

There may be some reason for why it was done this way that I'm not aware of but if you think what I've descibed makes sense, I'm happy to do a PR 😄

darkowlzz:
Addressed in #836

Issue #831, opened 2021-05-05
Consider adding support for bridge-less networking


networkop:
Hi 👋 , would you consider adding support for bridgeless networking using [tc redirect](https://github.com/awslabs/tc-redirect-tap), similar to what [kata containers](https://github.com/kata-containers/documentation/blob/master/design/architecture.md#networking) and [fircracker-containerd](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/docs/networking.md#a-preferred-traffic-control) are doing?
The use case is launching VMs either without any interface config or mounting/copying configuration files into the VM filesystem and activating them on boot with something like [ifreload](https://github.com/CumulusNetworks/ifupdown2), although ultimately it'll be down to the guest image to make sure this runs.

Implementation-wise, it could a simple check that in the [networking setup](https://github.com/weaveworks/ignite/blob/main/pkg/container/network.go#L62) for loop for any configured IP addresses. If there are more than 1, then use the existing bridge+dhcp approach; if not -- setup a tc redirect between the veth and tap links. 

I'd be happy to volunteer to do a PR, assuming this is deemed acceptable.

darkowlzz:
Implemented in #836

PR #836, opened 2021-05-10
Enable multiple non-IP interface to be connected via tc redirect

networkop:
Hi @darkowlzz @stealthybox 👋 
This is the PR that covers https://github.com/weaveworks/ignite/issues/832 and https://github.com/weaveworks/ignite/issues/831. At a high level it has the following impact:

### 1. Introduced a new CL argument `--sandbox-env-vars` accepting a comma-separated list of key=value pairs.
These values are passed to the respective container runtimes and used as env variables. I've had a choice two either create a new API version and add env vars as a new spec field or pass them around in VM's annotations. I've opted for the second option to minimise the impact of this change. I'm not sure if it's a good idea, happy to change it if necessary.

### 2. Introduced a new bool arg called `wait` to `StartVM` function - if set to false, this bypasses the `waitForSpawn` check.
This flag defaults to true for all existing function invocations to preserve backwards compatibility. However, when used via API, users can set this to false and skip the check for ignite-spawn. The purpose is to get the container PID to configure additional interfaces before ignite-spawn is fully initialised. 

### 3. Ignite-spawn can wait for a number of interfaces to be connected before firing up the VM. 
This is controlled through an environment variable called `IGNITE_INTFS`. To preserve backwards compatibility it defaults to 1, so without any variables set, the behaviour is the same as now. However, if this value is set to 1 on higher, `SetupContainerNetworking` will wait for that number of interfaces to be connected (up to a maximum timeout).

### 4. Ignite-spawn will connect additional veth and tap interfaces via tc redirect.
For backwards compatibility, the behaviour is to always use the current way of interconnecting interfaces (via bridge). However, if there's no IP on the interface, it will be interconnected with a VM via [tc redirect](https://github.com/kata-containers/kata-containers/blob/2fc7f75724ac9e18e60f63dcc9aa395dc51c184d/docs/design/architecture.md#networking). 

---

In general, all these changes strive to preserve the happy-path behavior of pre-existing code, so no major changes are expected for existing users.


cmd/ignite-spawn/ignite-spawn.go
```diff
@@ -96,3 +100,18 @@ func patchStopped(vm *api.VM) error {
 	patch := []byte(`{"status":{"running":false,"network":null,"runtime":null,"startTime":null}}`)
 	return patchutil.NewPatcher(scheme.Serializer).ApplyOnFile(constants.IGNITE_SPAWN_VM_FILE_PATH, patch, vm.GroupVersionKind())
 }
+
+func parseEnvVars(vars []string) map[string]string {
```
darkowlzz, line 104: Since this is used just once, above, maybe `os.Environ()` should be called in this function itself for now. Keeping true to the meaning of the function name.

cmd/ignite/run/create.go
```diff
@@ -95,6 +97,9 @@ func (cf *CreateFlags) NewCreateOptions(args []string, fs *flag.FlagSet) (*Creat
 		return nil, err
 	}
 
+	// Save env variables
+	baseVM.SetAnnotation(constants.IGNITE_ENV_VAR_ANNOTATION, cf.EnvVars)
```
darkowlzz, line 101: I think we need to provide a way to set this `cf.EnvVars` via the ignite configuration (global) as well, like other similar VM configurations in https://github.com/weaveworks/ignite/blob/999ad7c5ed35b8677a3950a0c227e78ee7d0956e/pkg/apis/ignite/types.go#L261-L266 to allow setting default for all the VMs and allow overriding that through `cf` flag and individual VM configuration.
Since this is set as annotation, the baseVM should have the annotation inherited from the ignite configuration. Annotation in VM configuration should override/append the baseVM configuration. The VM create flag value should append/override everything before.

cmd/ignite/run/start.go
```diff
@@ -78,7 +78,7 @@ func Start(so *StartOptions, fs *flag.FlagSet) error {
 		return err
 	}
 
-	if err := operations.StartVM(so.vm, so.Debug); err != nil {
+	if err := operations.StartVM(so.vm, so.Debug, true); err != nil {
```
darkowlzz, line 81: Maybe before setting true here, the VM configuration/annotation needs to be checked?

pkg/container/network.go
```diff
@@ -31,14 +35,34 @@ var ignoreInterfaces = map[string]struct{}{
 	"lo": {},
 }
 
-func SetupContainerNetworking() ([]DHCPInterface, error) {
+var (
+	maxIntfsVar     = "IGNITE_INTFS"
+	defaultMaxIntfs = 1
+)
+
+type FCInterface struct {
```
darkowlzz, line 43: A description about this interface would be nice to have in the godoc.

darkowlzz (commented):
Thanks for working on this.
Left an initial review with some suggestions.

darkowlzz:
> It seems like a number of e2e tests are failing, however they are also failing on the `main` branch. I'm not sure if anyone is aware of that?

We are aware of some flaky tests. But these look new failures, haven't seen them fail before.

networkop:
thanks @darkowlzz , I'm still working on this, so converted this to draft again. I'm also on PTO next week , so let's put this on hold until Monday week. I'll try to push out some updates by the 24th.

cmd/ignite-spawn/ignite-spawn.go
stealthybox: IMO using map of config key=values without mutating the ignite-spawn Environment is better, but this is conflating a lot of config ideas.  (flags producing ENV vs. flags overriding ENV)

In general though, is this the proper strategy for config?

Since we already have the information in the VM object, ignite-spawn has access to it and it can read the value from it directly.

That will get of all of the flag-passing and parsing code within ignite & ignite-spawn.

nit:
The names of the variables/annotations in this code should not reference any idea of ENV.
We should call it what it is with a proper domain.
There should be an explicit key for each thing you want.
```yaml
metadata:
  annotations:
    ignite.weave.works/interface-count: '3'
    # ignite.weave.works/interface-mode/eth0: 'dhcp-bridge'
    # ignite.weave.works/interface-mode/eth2: 'tc-redirect'
    # ignite.weave.works/interface-mode/macvtap-uvnrt: 'macvtap'
```

pkg/operations/start.go
```diff
@@ -80,6 +81,17 @@ func StartVM(vm *api.VM, debug bool) error {
...
+	// Only set envVars if at least one was configured
+	var envVars []string
+	parts := strings.Split(vm.GetAnnotation((constants.IGNITE_ENV_VAR_ANNOTATION)), ",")
+	for _, part := range parts {
+		if part == "" {
+			continue
+		}
+		envVars = append(envVars, part)
+	}
+	config.EnvVars = envVars
```
stealthybox, line 93: It's okay to remove this section since the VM Config file is already a sufficient mechanism to pass the config to ignite-spawn.

pkg/container/network.go
```diff
@@ -31,14 +35,34 @@ var ignoreInterfaces = map[string]struct{}{
...
+	VMTAP     string
+	MACFilter string
+}
+
+func SetupContainerNetworking(args map[string]string) ([]FCInterface, []DHCPInterface, error) {
```
stealthybox, line 48: Here instead of `map[string]string` containing custom-parse env, we can just parse annotations.
Alternatively, you could pass the entire VM object.
Ultimately, we will have a specific `VM.Spec.NetworkInfo` sub-type that could be passed here should the annotations graduate into a future API version.

stealthybox:
@networkop, thanks so much for working on this change. I'm really excited about it and you've inspired a really good use of annotations for low-cost experiments across ignite.

networkop:
brilliant, thanks for your comments @darkowlzz and @stealthybox. I think all of your suggestions make perfect sense and I'll make the necessary changes later this week. Hope to see y'all next Monday.

networkop:
Hi @darkowlzz @stealthybox , it was nice to chat to you today. Updated this PR, so most of your feedback got addressed. 
I think instead of re-writing the first comment I'll do a new PR summary below to preserve some history and context. 

This PR...

1. Introduced two new optional annotations:
    a. `ignite.weave.works/extra-intfs` -- can contain a comma-separated list of extra interfaces to be connected to the sandbox environment (by an external process, e.g. [containerlab](https://github.com/srl-labs/containerlab))
  b. `ignite.weave.works/sanbox-env` -- can contain a list of comma-separated key-value pairs to be passed into sandbox as environment variables. This would allow for firecracker timeouts to be adjusted, e.g. [like here].(https://github.com/networkop/containerlab/blob/ignite/runtime/ignite/ignite.go#L209)

2. Introduced a new bool arg called `wait` to `StartVM` function - if set to false, this bypasses the waitForSpawn check.
This flag defaults to true for all existing function invocations to preserve backwards compatibility. However, when used via API, users can set this to false and skip the check for ignite-spawn. The purpose is to get the container PID to configure additional interfaces before ignite-spawn is fully initialised.

3. Ignite-spawn can wait for a number of extra interfaces to be connected before firing up the VM.
This is controlled through the `ignite.weave.works/extra-intfs` annotation. If present, the value of this annotation is parsed as comma-separated list of strings where each string is expected to match one of the netlink interfaces inside a sandbox.  The VM will not start until all extra interfaces + main interface (eth0) are connected.

4. Ignite-spawn will connect additional veth and tap interfaces via tc redirect.
For backwards compatibility, the behavior is to always use the current way of interconnecting interfaces (via bridge). However, if there's no IP on the interface, it will be interconnected with a VM via tc redirect.


I still haven't touched the tests or documentation, so please let me know what you'd expect on that front.

networkop:
I had to make some extra changes to make e2e tests work. Mainly I had to expose the `wait` flag to allow start/run commands to allow a runtime to start in order to plug in additional interfaces.

networkop:
@stealthybox @darkowlzz I'm not sure why the tests are failing, they pass locally. I don't have the permissions to re-run, can you trigger a re-run please?

stealthybox:
> Introduced a new bool arg called wait to StartVM function - if set to false, this bypasses the waitForSpawn check.
This flag defaults to true for all existing function invocations to preserve backwards compatibility. However, when used via API, users can set this to false and skip the check for ignite-spawn. The purpose is to get the container PID to configure additional interfaces before ignite-spawn is fully initialised.

Talking through this with @darkowlzz, this has a subtle effect of changing the status start timestamp of a VM when the flag is disabled.

There are two approaches we came up with.

The one that is most similar to what you're wanting to do is to make `waitForSpawn` non-blocking and copy `StartVM()` to a new func called `StartVMNonBlocking()` which returns a done-channel.
This would allow you as an API client to do what you want to the VM while it's starting (like add interfaces), and it could pave the way for a general CLI option to run some start-lifecycle handlers.
In this case, for API compatibility, we would just wrap `StartVMNonBlocking` to keep the signature for the current blocking version.

The second option is to try and add formal lifecycle hooks that an API client could use.
This feels much harder and sort of out of scope.

WDYT?

pkg/operations/start.go
```diff
@@ -80,6 +81,17 @@ func StartVM(vm *api.VM, debug bool) error {
...
+	// Only set envVars if at least one was configured
+	var envVars []string
+	parts := strings.Split(vm.GetAnnotation((constants.IGNITE_SANDBOX_ENV_VAR)), ",")
+	for _, part := range parts {
+		if part == "" {
+			continue
+		}
+		envVars = append(envVars, part)
+	}
+	config.EnvVars = envVars
```
stealthybox, line 93: Could the API be a map instead of a CSV?
I'm slightly inspired by [docker-compose's `environment`](https://docs.docker.com/compose/compose-file/compose-file-v3/#environment).

ex:
```yaml
metadata:
  annotations:
    ignite.weave.works/sanbox-env/FIRECRACKER_GO_SDK_REQUEST_TIMEOUT_MILLISECONDS: '1000'
    ignite.weave.works/sanbox-env/FIRECRACKER_GO_SDK_INIT_TIMEOUT_SECONDS: '1000'  # did you mean '1' here?
```

Code diff would iterate through the annotation map and prefix match:
(untested)
```diff
diff --git pkg/constants/vm.go pkg/constants/vm.go
index fd45b7f9..a09021fa 100644
--- pkg/constants/vm.go
+++ pkg/constants/vm.go
@@ -42,6 +42,6 @@ const (
 	// IGNITE_EXTRA_INTFS is the annotation key to store a list of extra interfaces
 	IGNITE_EXTRA_INTFS = "ignite.weave.works/extra-intfs"
 
-	// IGNITE_SANDBOX_ENV_VAR is the annotation key to store a list of env variables
-	IGNITE_SANDBOX_ENV_VAR = "ignite.weave.works/sanbox-env"
+	// IGNITE_SANDBOX_ENV_PREFIX is the annotation prefix to store a map of env variables
+	IGNITE_SANDBOX_ENV_PREFIX = "ignite.weave.works/sanbox-env/"
 )
diff --git pkg/operations/start.go pkg/operations/start.go
index 08033c74..417d40f8 100644
--- pkg/operations/start.go
+++ pkg/operations/start.go
@@ -83,12 +83,11 @@ func StartVM(vm *api.VM, debug, wait bool) error {
 
 	// Only set envVars if at least one was configured
 	var envVars []string
-	parts := strings.Split(vm.GetAnnotation((constants.IGNITE_SANDBOX_ENV_VAR)), ",")
-	for _, part := range parts {
-		if part == "" {
-			continue
+	for anno, v := range vm.GetObjectMeta().Annotations {
+		if strings.HasPrefix(anno, constants.IGNITE_SANDBOX_ENV_PREFIX) {
+			k := strings.TrimPrefix(anno, constants.IGNITE_SANDBOX_ENV_PREFIX)
+			envVars = append(envVars, fmt.Sprintf("%s=%s", k, v))
 		}
-		envVars = append(envVars, part)
 	}
 	config.EnvVars = envVars
```

pkg/container/network.go
```diff
@@ -28,17 +35,31 @@ ip link set vm0 master br0
...
+	var fcIfaces firecracker.NetworkInterfaces
+
+	extraIntfs := parseExtraIntfs(vm)
+
+	// total number of interfaces is at least extraIntfs + eth0
+	totalIntfNum := len(extraIntfs) + 1
+
 	interval := 1 * time.Second
-	timeout := 1 * time.Minute
+	timeout := 2 * time.Minute
```
stealthybox, line 57: Just noting that the default network-setup retry-timeout has been increased. :+1:

pkg/container/network.go
```diff
@@ -53,35 +74,82 @@ func SetupContainerNetworking() ([]DHCPInterface, error) {
...
+	if err != nil || allIfaces == nil || len(allIfaces) == 0 {
 		return true, fmt.Errorf("cannot get local network interfaces: %v", err)
 	}
 
-	// interfacesCount counts the interfaces that are relevant to Ignite (in other words, not ignored)
-	interfacesCount := 0
+	ifaces := filterIgnored(allIfaces, extraIntfs)
+	if len(ifaces) < *expectedIntfNum {
+		return true, fmt.Errorf("not enough extra interfaces connected (%d/%d), waiting", len(ifaces), *expectedIntfNum)
+	}
```
stealthybox, line 117: If we add `vm` to the `networkSetup` function signature, we can get rid of all of the counting logic and the passing around of a single number.

It will be possible here to just loop through the interface names and compare them with what's in the netns.

This will clean up the code.
The error message will be able to include exactly which interface names are missing and what their expected modes(`tc-redirect`/`dhcp-bridge`/`macvtap`) are.

pkg/container/network.go
```diff
@@ -53,35 +74,82 @@ func SetupContainerNetworking() ([]DHCPInterface, error) {
...
-		// Skip the interface if it's ignored
-		if _, ok := ignoreInterfaces[iface.Name]; ok {
+
+		// Try to transfer the address from the container to the DHCP server
+		ipNet, gw, noIPs, err := takeAddress(&iface)
+
+		// If interface has no IPs configured, setup tc redirect
+		if noIPs && iface.Name != mainInterface {
+			log.Printf("Interface %s has no IP, setting up tc redirect", iface.Name)
+			tcInterface, err := addTcRedirect(&iface)
```
stealthybox, line 132: Maybe an explicit opt-in to the tc-redirect interface mode can eliminate the need for a heuristic check here.
It may make it easier to differentiate from `macvtap` and support future, esoteric, interface modes.

Does that sound like a good decision to you?

Example API from https://github.com/weaveworks/ignite/pull/836#discussion_r633692902:
```yaml
metadata:
  annotations:
    ignite.weave.works/interface-mode/eth0: 'dhcp-bridge'
    ignite.weave.works/interface-mode/eth2: 'tc-redirect'
    ignite.weave.works/interface-mode/eth3: 'tc-redirect'
    ignite.weave.works/interface-mode/macvtap-uvnrt: 'macvtap'
    ignite.weave.works/interface-mode/macvtap-hgfeq: 'macvtap'
```

The key format can be whatever makes sense to you.
`ignite.weave.works/interface/eth1` or `ignite.weave.works/interface/eth1/mode` could be good alternatives.
This allows us to add expected behavior about each interface without parsing a custom CSV+K=V format.

Parsing would be a similar iteration with prefix/suffix filtering, similar to the sandbox-env suggestion.

networkop:
> > Introduced a new bool arg called wait to StartVM function - if set to false, this bypasses the waitForSpawn check.
> > This flag defaults to true for all existing function invocations to preserve backwards compatibility. However, when used via API, users can set this to false and skip the check for ignite-spawn. The purpose is to get the container PID to configure additional interfaces before ignite-spawn is fully initialised.
> 
> Talking through this with @darkowlzz, this has a subtle effect of changing the status start timestamp of a VM when the flag is disabled.
> 
> There are two approaches we came up with.
> 
> The one that is most similar to what you're wanting to do is to make `waitForSpawn` non-blocking and copy `StartVM()` to a new func called `StartVMNonBlocking()` which returns a done-channel.
> This would allow you as an API client to do what you want to the VM while it's starting (like add interfaces), and it could pave the way for a general CLI option to run some start-lifecycle handlers.
> In this case, for API compatibility, we would just wrap `StartVMNonBlocking` to keep the signature for the current blocking version.

Thanks @stealthybox , I just want to make sure I understand what you're saying.  I have two different interpretations:

First one:
1. Add two new functions `StartVMBlocking` and `StartVMnonBlocking`, each one calling `StartVM` with the `wait` bool either on or off.
2. Modify `StartVM` return signature to `(error, <-chan struct{}) `, the latter can be used to check `waitForSsh`
3. In the CLI, depending on the `wait` flag invoke either one of these functions

Second one:
1. Leave the `StartVM()` function as it was before, add a new function `StartVMNonBlocking()` with the same code but returning (error, <-chan bool) instead.
2. In the CLI, depending on the `wait` flag invoke one or the other.

pkg/operations/start.go
networkop: yep, got it

pkg/container/network.go
networkop: yep, found that my old i5 was struggling with 2 VMs in 1 minute

pkg/container/network.go
networkop: ok, yeah, makes sense.

networkop:
@darkowlzz @stealthybox PR is ready for review.

pkg/container/network.go
```diff
@@ -293,3 +488,33 @@ func maskString(mask net.IPMask) string {
...
+			continue
+		}
+
+		switch v {
+		case supportedModes.dhcp:
+			result[k] = v
+		case supportedModes.tc:
+			result[k] = v
+		default:
+			continue
```
stealthybox, line 513: Should we print a warning for unsupported mode here?
Ideally, it will show up in the ignite spawn log.

stealthybox:
```diff
commit 219c8e89f2b80373cfbbf53de6c21101f24cb672
Author: leigh capili <leigh@null.net>
Date:   Mon Jun 21 10:34:48 2021 -0600

    Use set for interface modes

diff --git pkg/container/network.go pkg/container/network.go
index 2c28e99c..af00e47f 100644
--- pkg/container/network.go
+++ pkg/container/network.go
@@ -33,18 +33,20 @@ ip link set eth0 master br0
 ip link set vm0 master br0
 */
 
-// Array of container interfaces to ignore (not forward to vm)
+// ignoreInterfaces is a Set of container interface names to ignore (not forward to vm)
 var ignoreInterfaces = map[string]struct{}{
 	"lo":    {},
 	"sit0":  {},
 	"tunl0": {},
 }
 
-var supportedModes = struct {
-	dhcp, tc string
-}{
-	dhcp: "dhcp-bridge",
-	tc:   "tc-redirect",
+const MODE_DHCP = "dhcp-bridge"
+const MODE_TC = "tc-redirect"
+
+// supportedModes is a Set of Mode strings
+var supportedModes = map[string]struct{}{
+	MODE_DHCP: {},
+	MODE_TC:   {},
 }
 
 var mainInterface = "eth0"
@@ -57,7 +59,7 @@ func SetupContainerNetworking(vm *api.VM) (firecracker.NetworkInterfaces, []DHCP
 
 	// Setting up mainInterface if not defined
 	if _, ok := vmIntfs[mainInterface]; !ok {
-		vmIntfs[mainInterface] = supportedModes.dhcp
+		vmIntfs[mainInterface] = MODE_DHCP
 	}
 
 	interval := 1 * time.Second
@@ -116,7 +118,7 @@ func collectInterfaces(vmIntfs map[string]string) (bool, error) {
 		// default fallback behaviour to always consider intfs with an address
 		addrs, _ := intf.Addrs()
 		if len(addrs) > 0 {
-			vmIntfs[intf.Name] = supportedModes.dhcp
+			vmIntfs[intf.Name] = MODE_DHCP
 		}
 	}
 
@@ -127,7 +129,7 @@ func collectInterfaces(vmIntfs map[string]string) (bool, error) {
 		}
 
 		// for DHCP interface, we need to make sure IP and route exist
-		if mode == supportedModes.dhcp {
+		if mode == MODE_DHCP {
 			intf := foundIntfs[intfName]
 			_, _, _, noIPs, err := getAddress(&intf)
 			if err != nil {
@@ -163,7 +165,7 @@ func networkSetup(fcIntfs *firecracker.NetworkInterfaces, dhcpIntfs *[]DHCPInter
 		}
 
 		switch vmIntfs[intfName] {
-		case supportedModes.dhcp:
+		case MODE_DHCP:
 			ipNet, gw, err := takeAddress(intf)
 			if err != nil {
 				return fmt.Errorf("error parsing interface %q: %s", intfName, err)
@@ -185,7 +187,7 @@ func networkSetup(fcIntfs *firecracker.NetworkInterfaces, dhcpIntfs *[]DHCPInter
 					HostDevName: dhcpIface.VMTAP,
 				},
 			})
-		case supportedModes.tc:
+		case MODE_TC:
 			tcInterface, err := addTcRedirect(intf)
 			if err != nil {
 				log.Errorf("Failed to setup tc redirect %v", err)
@@ -494,22 +496,20 @@ func maskString(mask net.IPMask) string {
 func parseExtraIntfs(vm *api.VM) map[string]string {
 	result := make(map[string]string)
 
-	for k, v := range vm.GetObjectMeta().Annotations {
-		if !strings.HasPrefix(k, constants.IGNITE_INTERFACE_ANNOTATION) {
+	for intf, mode := range vm.GetObjectMeta().Annotations {
+		if !strings.HasPrefix(intf, constants.IGNITE_INTERFACE_ANNOTATION) {
 			continue
 		}
 
-		k = strings.TrimPrefix(k, constants.IGNITE_INTERFACE_ANNOTATION)
-		if k == "" {
+		intf = strings.TrimPrefix(intf, constants.IGNITE_INTERFACE_ANNOTATION)
+		if intf == "" {
 			continue
 		}
 
-		switch v {
-		case supportedModes.dhcp:
-			result[k] = v
-		case supportedModes.tc:
-			result[k] = v
-		default:
+		if _, ok := supportedModes[mode]; ok {
+			result[intf] = mode
+		} else {
+			log.Warnf("VM specifies unrecognized mode %q for interface %q", mode, intf)
 			continue
 		}
 

```

pkg/container/network.go
```diff
@@ -28,17 +35,40 @@ ip link set vm0 master br0
...
+
 	interval := 1 * time.Second
-	timeout := 1 * time.Minute
+	timeout := 2 * time.Minute
 
 	err := wait.PollImmediate(interval, timeout, func() (bool, error) {
+
 		// This func returns true if it's done, and optionally an error
-		retry, err := networkSetup(&dhcpIfaces)
+		retry, err := collectInterfaces(vmIntfs)
```
stealthybox, line 69: noting here, this will always have `eth0` as the main interface included here

pkg/operations/start.go
```diff
@@ -177,9 +207,10 @@ func waitForSpawn(vm *api.VM) error {
 		time.Sleep(checkInterval)
 
 		if util.FileExists(path.Join(vm.ObjectPath(), constants.PROMETHEUS_SOCKET)) {
-			return nil
+			vmChans.SpawnFinished <- nil
+			return
 		}
 	}
 
-	return fmt.Errorf("timeout waiting for ignite-spawn startup")
```
stealthybox, line 184: Please move https://github.com/weaveworks/ignite/pull/836/files#diff-fe685b5ad77268d55269c38f483185a644e6b741eddc7b86511bbe3d51fe0ad7L133-L136 here and call `providers.Client.VMs().Set(vm)` one more time.

```go
	// Set the start time for the VM
	startTime := apiruntime.Timestamp()
	vm.Status.StartTime = &startTime
	vm.Status.Network.Plugin = providers.NetworkPluginName
```

Goal is for waitForSpawn to update the VM object with the proper VM start time.

stealthybox:
```diff
diff --git pkg/operations/start.go pkg/operations/start.go
index b553fc3c..06d868a1 100644
--- pkg/operations/start.go
+++ pkg/operations/start.go
@@ -153,30 +153,28 @@ func StartVMNonBlocking(vm *api.VM, debug bool) (*VMChannels, error) {
 		log.Infof("Started Firecracker VM %q in a container with ID %q", vm.GetUID(), containerID)
 	}
 
-	// TODO: This is temporary until we have proper communication to the container
-	go waitForSpawn(vm, vmChans)
-
 	// Set the container ID for the VM
 	vm.Status.Runtime.ID = containerID
 	vm.Status.Runtime.Name = providers.RuntimeName
 
-	// Set the start time for the VM
-	startTime := apiruntime.Timestamp()
-	vm.Status.StartTime = &startTime
-	vm.Status.Network.Plugin = providers.NetworkPluginName
-
 	// Append non-loopback runtime IP addresses of the VM to its state
 	for _, addr := range result.Addresses {
 		if !addr.IP.IsLoopback() {
 			vm.Status.Network.IPAddresses = append(vm.Status.Network.IPAddresses, addr.IP)
 		}
 	}
+	vm.Status.Network.Plugin = providers.NetworkPluginName
+
+	// write the API object in a non-running state before we wait for spawn's network logic and firecracker
+	if err := providers.Client.VMs().Set(vm); err != nil {
+		return vmChans, err
+	}
 
-	// Set the VM's status to running
-	vm.Status.Running = true
+	// TODO: This is temporary until we have proper communication to the container
+	// It's best to perform any imperative changes to the VM object pointer before this go-routine starts
+	go waitForSpawn(vm, vmChans)
 
-	// Write the state changes
-	return vmChans, providers.Client.VMs().Set(vm)
+	return vmChans, nil
 }
 
 // verifyPulled pulls the ignite-spawn image if it's not present
@@ -202,12 +200,26 @@ func waitForSpawn(vm *api.VM, vmChans *VMChannels) {
 	const timeout = 10 * time.Second
 	const checkInterval = 100 * time.Millisecond
 
-	startTime := time.Now()
-	for time.Since(startTime) < timeout {
+	timer := time.Now()
+	for time.Since(timer) < timeout {
 		time.Sleep(checkInterval)
 
 		if util.FileExists(path.Join(vm.ObjectPath(), constants.PROMETHEUS_SOCKET)) {
-			vmChans.SpawnFinished <- nil
+			// Before we write the VM, we should REALLY REALLY re-fetch the API object from storage
+			vm, err := providers.Client.VMs().Get(vm.GetUID())
+			if err != nil {
+				vmChans.SpawnFinished <- err
+			}
+
+			// Set the start time for the VM
+			startTime := apiruntime.Timestamp()
+			vm.Status.StartTime = &startTime
+
+			// Set the VM's status to running
+			vm.Status.Running = true
+
+			// Write the state changes, send any errors through the channel
+			vmChans.SpawnFinished <- providers.Client.VMs().Set(vm)
 			return
 		}
 	}
```

e2e/multinet_test.go
```diff
@@ -0,0 +1,323 @@
...
+	// Get the VM ID
+	idCmd := igniteCmd.New().
+		With("ps").
+		With("--filter").
+		With(fmt.Sprintf("{{.ObjectMeta.Name}}=%s", vmName)).
+		With("--quiet")
+
+	idOut, idErr := idCmd.Cmd.CombinedOutput()
+	assert.Check(t, idErr, fmt.Sprintf("vm id not found: \n%q\n%s", idCmd.Cmd, idOut))
+	vmID := string(idOut[:len(idOut)-2])
```
stealthybox, line 264: @darkowlzz suggests doing something similar to `TestRunWithoutName` but for the VM ID instead of slicing the stdout:
```go
		With("--template={{.ObjectMeta.Name}}")
```

networkop:
@darkowlzz @stealthybox  I'm running into an issue after our latest `waitForSpawn` refactoring. Since we moved the `vm.Status.Running = true` into a goroutine, I can no longer ignore the vmChannels , i.e. I can't do:
```go
_, err = operations.StartVMNonBlocking(so.vm, so.Debug)
```
because the VM status never changes to running and from the pov of ignite it remains in the following state:

```bash
VM ID                   IMAGE                           KERNEL                                  SIZE    CPUS    MEMORY          CREATED      STATUS          IPS             PORTS   NAME
a3710f7f8f76395e        weaveworks/ignite-ubuntu:latest weaveworks/ignite-kernel:5.4.108        3.0 GB  1       800.0 MB        32s ago      *Up <nil>       192.168.223.2           e2e-test-vm-multinet
WARN[0000] The symbol * on the VM status indicates that the VM manifest on disk may not be up-to-date with the actual VM status from the container runtime
```

I'm not sure what's the best way forward now. On the one hand, we can't move `Running=true` before `waitForSpawn` because the VM is not started, on the other hand, I need to return from the `StartVMNonBlocking` in order to connect additional interfaces. 

WDYT?

networkop:
@darkowlzz @stealthybox can you have a look at the latest commit? I've re-worked one of the e2e tests to use ignite API instead of CLI. This would allow us to drop the `--wait` CLI flag (because of the above reasons) and wait for the vmChannels to return before doing `exec`. 
If you think it looks alright, I'll update the remaining two tests in a similar way.

stealthybox:
@networkop I think this is a good way forward for the tests 👍

Perhaps we could copy the waitForSpawn code into containerlab?
I think the better structural change is to support the channels in containerlab.

networkop:
cool, I think this is it then, we've got all the changes from yesterday merged. two major deviations are
* dropped the `--wait` flag in CLI
* refactored e2e tests to use API

what do y'all think?

e2e/multinet_test.go
```diff
@@ -9,72 +9,105 @@ import (
...
+)
+
+func startAsyncVM(t *testing.T, intfs []string) (*operations.VMChannels, string) {
+
 	assert.Assert(t, e2eHome != "", "IGNITE_E2E_HOME should be set")
+	igniteUtil.GenericCheckErr(providers.Populate(ignite.Preload))
 
-	vmName := "e2e-test-vm-multinet"
+	_ = igniteDocker.SetDockerRuntime()
+	_ = igniteDocker.SetDockerNetwork()
```
stealthybox, line 37: It's a natural consequence of globals being side-effected here that these tests cannot be parallelized with other test runs that need different values.
This doesn't affect the existing test suites which make individual invocations of the ignite binary.
Also, we already run our e2e suite in serial anyway.

All of this is fine, but it's worth noting. :+1:

e2e/multinet_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	dockerCmd.New().
+		With("exec", fmt.Sprintf("ignite-%s", vmID)).
+		With("ip", "link", "set", "foo", "address", fooAddr).
+		Run()
+
+	// check that the VM has started before trying exec
+	if err := <-vmChans.SpawnFinished; err != nil {
+		t.Fatalf("failed to start a VM: \n%q\n", err)
+	}
```
stealthybox, line 123: I like that these tests exercise the wait code that polls for the Firecracker Prometheus socket.

stealthybox (approved):
LGTM

@darkowlzz should probably review as well considering the new e2e technique

@networkop do you think you can make containerlab work with the channels?

networkop:
@stealthybox maybe, I had an idea of adding an extra func to clab's runtime interface, something like `waitForWarmup` that will make sure all of the individual containers have started. I need to see how it will work with the rest of the runtimes.

cmd/ignite/run/exec.go
```diff
@@ -30,5 +31,8 @@ func (ef *ExecFlags) NewExecOptions(vmMatch string, command ...string) (eo *Exec
 
 // Exec executes command in a VM based on the provided ExecOptions.
 func Exec(eo *ExecOptions) error {
+	if err := waitForSSH(eo.vm, constants.SSH_DEFAULT_TIMEOUT_SECONDS, 5); err != nil {
```
darkowlzz, line 34: This may not be a concern, but I noticed that `runSSH()` below accepts a timeout parameter. If this `waitForSSH()` with default timeout of 5 seconds runs before `runSSH()`, the user provided timeout will not work, `waitForSSH()` will return error after waiting for 5 seconds and `runSSH()` with user provided timeout will never be executed.
Should the timeout for `waitForSSH()` be the same as provided in the `ExecOptions.Timeout`?

darkowlzz (approved):
Overall, looks good to me.
Left a comment about the wait in exec command which may not be concerning.

cmd/ignite/run/exec.go
networkop: oh yeah, that's a good catch. I've added it there to workaround the behavior introduced by the `--wait` flag and I forgot to remove it. Will fix this up.

networkop: I think my above conclusion was wrong. Complete removal of `waitForSSH` leads to multinet tests failure when the prometheus socket is up -> VM is started but SSH is still not up yet. So we need it there, but I've changed the 5 second timeout to the user-provided timeout value.

stealthybox:
Nice
Shall I purchase some ice-cream cake to celebrate?

networkop:
brilliant! thanks @darkowlzz and @stealthybox for all the support 🙏

PR #849, opened 2021-06-28
kernel config updates

networkop:
This change enables some features required to simulate multi-homed end hosts using standard weave kernel.

* Bonding & Teaming -- enable link aggregation
* IPv6 multiple tables -- enables routing policy database for IPv6 (it's already enabled for IPv4).

This PR also bumps minor kernel versions


darkowlzz (approved):
LGTM!
Thanks for updating all the kernel versions.

PR #879, opened 2021-10-24
Fix #872, Adding Almalinux and Rocky Linux support

srbala:
Replacing PR #876, to keep history clean.

Adding Almalinux and Rocky Linux support, Resolves Fix #872

PR #882, opened 2021-11-05
Use portable sed inlines

stealthybox:
ref: https://unix.stackexchange.com/a/92907
Related discussion: https://github.com/weaveworks/ignite/pull/879#issuecomment-954912753


darkowlzz (approved):
LGTM!

PR #863, opened 2021-08-12
Bump CNI Plugins to v1.0.1

gaby:
CNI Plugins have officially graduated to stable with version 1.0.1 released yesterday. The biggest breaking change is the removal of the flannel plugin.

Release notes here: https://github.com/containernetworking/plugins/releases/tag/v1.0.1

PR #903, opened 2022-03-07
update cni version to 1.0.1

abn:
This builds on top of the work done by @gaby in #863.

Closes: #863


gaby:
👍🏻 For this, I couldnt get my builds pass for #863

gaby:
@chanwit @darkowlzz @stealthybox any updates on this?

darkowlzz:
Hi @abn and @gaby , sorry for the late response. Before we update CNI, we'll have to bring back ignite to a good shape.
I've created https://github.com/weaveworks/ignite/pull/923 to fix the current CI failures.
I also see a lot of dependency updates in this PR that may not be directly related to CNI. I can try updating those dependencies in a separate PR once the builds are in a good shape. Just trying to reduce the change surface to avoid any unintended changes.

Thanks for pushing this forward.

darkowlzz:
Hi, I tried updating the unrelated dependencies first and when I updated containerd, it also updates CNI to v1, which also required updating firacracker-go-sdk, as the old version no longer worked with new CNI, and everything else. So at the end, I ended up with everything that's in this PR.
The main branch is in a good state now with all the pipelines passing. It'd be good if you just rebase this and see how the pipeline behaves.

I'd like to know the reason for changing `run` to `create` in e2e test `e2e/config_test.go` in the last commit. I believe we need it to be `run` for the runtime to be set on the VMs. Create just creates the object on disk. Since the CI is green now, maybe you can try to see what happens without those change?

go.mod
```diff
@@ -47,13 +38,13 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	github.com/stretchr/testify v1.6.1
-	github.com/vishvananda/netlink v1.1.0
+	github.com/stretchr/testify v1.7.0
+	github.com/vishvananda/netlink v1.1.1-0.20220224135417-f7fd7af4377d
```
darkowlzz, line 42: This can be `v1.2.0-beta` now.

PR #905, opened 2022-03-17
fix wget: "error getting response" problem

Rickylss:
While BASEIMAGE is alpine and the default wget
doesn't work with HTTPS. We need update both
openssl and wget.

Signed-off-by: haibiao.xiao <xiaohaibiao331@outlook.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,11 @@ FROM BASEIMAGE AS build
 COPY qemu-QEMUARCH-static /usr/bin/
 
 # device-mapper is needed for snapshot functionalities
+# openssl and wget needes to be updated.
 RUN apk add --no-cache \
-    device-mapper
+    device-mapper \
+    openssl \
+    wget
 
 # Download the Firecracker binary from Github
 ARG FIRECRACKER_VERSION


Dockerfile
```diff
@@ -5,8 +5,11 @@ FROM BASEIMAGE AS build
 COPY qemu-QEMUARCH-static /usr/bin/
 
 # device-mapper is needed for snapshot functionalities
+# openssl and wget needes to be updated.
 RUN apk add --no-cache \
-    device-mapper
+    device-mapper \
+    openssl \
+    wget
```
darkowlzz, line 12: This Dockerfile is for the ignite sandbox container image. Can you share what's the use case of wget in the sandbox?
If it's a problem in the OS images, we can do something similar in the OS images.

Rickylss: image
In step 5, we use wget pull the firecracker image, and we get an error.

image
And in the docker container, we can wget http but not https. Then we update wget by `apk add  wget` and wget works on https now.

So, it's actually a problem in the OS images.

Giles2000: i have a program same here,but is a little different is the error code is 'time out',
so it's the build container can‘t connect the github.
i have a VPN in my pc , so change the network mode to "host" in the build container by recode the Makefile file. it's worked

```suggestion
 @@ -133,7 +134,8 @@ else
 endif
        $(DOCKER) build -t $(IMAGE):${IMAGE_DEV_TAG}-$(GOARCH) \
                --build-arg FIRECRACKER_VERSION=${FIRECRACKER_VERSION} \
-               --build-arg FIRECRACKER_ARCH_SUFFIX=${FIRECRACKER_ARCH_SUFFIX} bin/$(GOARCH)
+               --build-arg FIRECRACKER_ARCH_SUFFIX=${FIRECRACKER_ARCH_SUFFIX} bin/$(GOARCH)\
+               --network host

```
all in all, solve the network program. there are a lot of ways ,use the method you are most familiar with

PR #922, opened 2022-04-29
fix match error image program

Giles2000:
there is a image named app_log:v22 has been imported.
now i want to use a new image which name is app_log:v2 to create a vm
ignite will use the wrong image

[root@yg_x86 hostname]# ignite images
4f8db4f129f02838        10.20.33.115/library/app_log:v22                                        38s     16.8 MB

[root@yg_x86 hostname]# ignite create  10.20.33.115/library/app_log:v2
"without pulling"
INFO[0001] Created VM with ID "f9f07daad5195ce7" and name "spring-dew" 

[root@yg_x86 hostname]# ignite inspect vm f9f07daad5195ce7
{
  ...
  "spec": {
    "image": {
      "oci": "10.20.33.115/library/app_log:v22"


darkowlzz:
Hi @Giles2000 , thanks for submitting a fix for this issue. Looks like the bug is in a library we use https://github.com/weaveworks/libgitops/ . We can't update the code inside vendor directory directly. Can you please submit a fix in libgitops project? Once that's fixed in upstream, we can pull the new libgitops and fix it in ignite.

I was able to reproduce it easily by just importing images that have similar names:
```console
$ docker tag weaveworks/ignite-ubuntu:20.04-amd64 test/ubuntu:v22
$ docker tag weaveworks/ignite-ubuntu:20.04-amd64 test/ubuntu:v2

# ignite image import test/ubuntu:v22 --runtime docker
INFO[0000] Starting image import...
INFO[0002] Imported OCI image "test/ubuntu:v22" (218.8 MB) to base image with UID "b3e740e5319958cd"
INFO[0002] Created image with ID "b3e740e5319958cd" and name "test/ubuntu:v22"

# ignite image import test/ubuntu:v2 --runtime docker
INFO[0000] Created image with ID "b3e740e5319958cd" and name "test/ubuntu:v22"
```

Giles2000:
@darkowlzz  ok

Giles2000:
hi @darkowlzz 
em...we have a program.
at commit [c4658](https://github.com/weaveworks/libgitops/commit/c46586bf9637cdf7283b0a378b5c9172f9d0891a#) in [lib](https://github.com/weaveworks/libgitops/ .) . it has removed idname.go after releases v0.0.2.
i think someone contact then to revisit idname.go and fix the bug is a better way becouse they write 'Revisit if we need this file/package in the future.' in new idname.go,
Otherwise, we hava to  rebuild the code which use the function in idname.go? it's a hard work.
can you contact them to revisit it ？ because you're the same roomMember of Weaveworks, maybe it's more effectively

darkowlzz:
@Giles2000 thanks for sharing the details. So ignite is using an older version of libgitops. Maybe we'll have to update to the new version and change the ignite code accordingly.
I'll have to check the details and see what can be done.

PR #923, opened 2022-04-30
Update go to 1.17.9

darkowlzz:
Updates go to 1.17.9 and golangci-lint to v1.45.2.

Also replaces `golang.org/x/crypto/ssh/terminal` with `golang.org/x/term` due to golangci-lint static check error:
```
SA1019: package golang.org/x/crypto/ssh/terminal is deprecated: this package moved to golang.org/x/term.
```


stealthybox:
Thanks so much for this maintenance cleanup!

PR #924, opened 2022-04-30
e2e: Set containerd socket address in RMI command

darkowlzz:
In machines where continerd socket isn't in
/run/containerd/containerd.sock, the default path, the tests fail
to cleanup the downloaded images in registry auth tests, resulting in
unexpected test failures due to old images.

Export StatContainerdSocket() from pkg/runtime/containerd/ and use it in
the e2e test to determine the containerd socket.

The test failures are unrelated and are fixed by https://github.com/weaveworks/ignite/pull/923 .

### DIFF
--- a/e2e/util/image.go
+++ b/e2e/util/image.go
@@ -4,6 +4,7 @@ import (
 	"os/exec"
 
 	"github.com/weaveworks/ignite/pkg/runtime"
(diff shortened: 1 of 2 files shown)


stealthybox:
Ooh nice catch. That function makes sense to export, and I'm glad it's already factored nicely, where it doesn't require any changes to do so.

PR #936, opened 2022-06-10
add parameter judgment for "--cpus num"

slark-yuxj:
add parameter judgment for "--cpus num". When num does not meet the conditions (1 or even numbers between 1 and 32), you need to report an error and do not create a virtual machine


stealthybox:
Thanks for the patch!

Curious, where did you get a max count of 32 from?
I know there is hardware that has many more cores than that.

PR #938, opened 2022-06-28
Rebase alpine image at edge tar file

Sfinx:
Seems like stock alpine FC ext4 image is rather old. Rebase it using edge tar from alpine project


gaby:
@Sfinx Why did you remove the other kernel versions other than 5.18? Those are used by ignite.

Sfinx:
> @Sfinx Why did you remove the other kernel versions other than 5.18? Those are used by ignite.

Ignore 87430c8 - it do not related. Only cff1eba is

